PR #1, opened 2018-03-14
COMPASS-2636: generic test runner to test lang to lang conversion

lrlna:
This PR has a test that takes in a json file and makes sure the language to language code generation is as expected.


test/language-conversion.json
```diff
@@ -0,0 +1,100 @@
...
+      "java": ".append(\"nl\", new java.lang.Long(3))",
+      "csharp": ""
+    },
+    "ni": {
+      "javascript": "new Number(2)",
+      "python": "int(2)",
+      "java": ".append(\"ni\", new java.lang.Integer(2))",
+      "csharp": ""
+    },
+    "dec128": {
```
aherlihy: TODO: Should test the `Decimal128.toString` util as well.

test/language-conversion.json
```diff
@@ -0,0 +1,100 @@
...
+      "java": ".append(\"minKey\", new MinKey())",
+      "csharp": ""
+    },
+    "maxkey": {
+      "javascript": "MaxKey()",
+      "python": "MaxKey()",
+      "java": ".append(\"maxKey\", new MaxKey())",
+      "csharp": ""
+    },
+    "isodate": {
```
aherlihy: TODO: test a reasonable number of date constructors

test/language-conversion.json
```diff
@@ -0,0 +1,100 @@
...
+      "javascript": "new Date('1995-12-17T03:24:00')",
+      "python": "datetime.date(1995, 11, 17)",
+      "java": ".append(\"isodate\", new java.util.Date(1995, 11, 17))",
+      "csharp": ""
+    },
+    "regexp": {
```
aherlihy: We have to generate JS RegExps but there is _also_ a BSONRegExp class that I didn't realize existed: https://github.com/mongodb/js-bson/blob/1.0-branch/lib/bson/regexp.js

test/language-conversion.json
```diff
@@ -0,0 +1,100 @@
...
+      "javascript": "new RegExp('/^[a-z0-9_-]{3,16}$/')",
+      "python": "RegExp('/^[a-z0-9_-]{3,16}$/')",
+      "java": ".append(\"regexp\", Pattern.compile(\"/^[a-z0-9_-]{3,16}$/\"))",
+      "csharp": ""
+    },
+    "ts": {
+      "javascript": "Timestamp(10, 100)",
```
aherlihy: TODO: lots of built-in methods/utils

test/language-conversion.json
```diff
@@ -0,0 +1,100 @@
...
+      "python": "Timestamp(10, 100)",
+      "java": ".append(\"ts\", new BSONTimestamp(10, 100))",
+      "csharp": ""
+    },
+    "subdoc": {
```
aherlihy: Should add tests for further nested documents, and documents that contain arrays.

test/language-conversion.json
```diff
@@ -0,0 +1,100 @@
...
+      "javascript": "{x: 1}",
+      "python": "{'x': 1}",
+      "java": ".append(\"subDoc\", new Document(\"x\", 1))",
+      "csharp": ""
+    },
+    "subarray": {
```
aherlihy: Should add tests for further nested arrays, and arrays that contain documents.

test/language-conversion.json
```diff
@@ -0,0 +1,100 @@
+{
+  "languages": [
+    "javascript",
+    "python",
+    "java",
+    "csharp"
+  ],
+  "bsonTypes": {
+    "code": {
```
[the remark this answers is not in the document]
lrlna: hmmmm there will be a lot of instances of similar situations where we test all or partial arguments given to a function. we can just add them on as new objects in the `bsonTypes` object i think as we come up with something new.

test/language-conversion.json
```diff
@@ -0,0 +1,100 @@
...
+    "dbref": {
+      "javascript": "DBRef('collection', 'objectID', 'dbString')",
+      "python": "DBRef('collection', 'objectID', 'dbString')",
+      "java": "",
+      "csharp": ""
+    },
+    "nd": {
+      "javascript": "Double(3)",
+      "python": "float(3)",
+      "java": ".append(\"nd\", new java.lang.Double(3.00))",
```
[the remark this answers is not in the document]
lrlna: oh yea i was going to ask, is java going to be outputted as the entire `.append()` thing or just `new java`... cause yea, it's weird to have the append thing

test/language-conversion.json
```diff
@@ -0,0 +1,148 @@
...
+    "code": {
+      "javascript": "Code('some code')",
+      "python": "Code('some code')",
+      "java": "new Code(\"some code\")",
+      "csharp": ""
+    },
+    "codeWithScope": {
+      "javascript": "Code('some code', {})",
+      "python": "Code('some code', {})",
+      "java": "new Code(\"some code\", new Document())",
```
aherlihy: This is going to be `CodeWithScope` for java

test/language-conversion.json
```diff
@@ -0,0 +1,148 @@
...
+      "csharp": ""
+    },
+    "integerLiteral": {
+      "javascript": "2",
+      "python": "2",
+      "java": "2",
+      "csharp": ""
+    },
+    "dec128": {
+      "javascript": "new Decimal128.fromString('5.00000001')",
```
aherlihy: We should just call the constructor here, using `bytes`: https://github.com/mongodb/js-bson/blob/1.0-branch/lib/bson/decimal128.js#L181. The `Decimal128.fromString` is a build-in util that we'll implement later on.

test/bson-constructor.test.js
```diff
@@ -0,0 +1,39 @@
...
+
+const parseResult = parse(fs.readFileSync(path.join(__dirname, './language-conversion.json')));
+// if an error is returned from parsing json, just throw it
+if (parseResult.err) throw new Error(parseResult.err.message);
+const constructorJSON = parseResult.value;
+
+const languages = constructorJSON.languages;
+const bsonTypes = constructorJSON.bsonTypes;
+
+languages.forEach((lang) => {
```
aherlihy: The only tricky thing with this is that javascript input vs. output is not exactly the same. For example, if you input into javascript `Binary("test")`, it will output always as `Binary("test", Binary.DEFAULT_TYPE)`. 

Right now, the javascript input and output are roughly identical, but the issue will come up when we expand the input language to support shell syntax. We still have to do a little bit of transpiling for a javascript-to-javascript conversion because it's actually "shell javascript" to "driver javascript". 

I think we probably just want to add another field that we set on the JSON object called "input" so that we can separate out input JS from output JS.

lrlna: I fixed it to be an array of `inputLanguages` in the json and later on, when it comes to it, we can add an output language array and map the two together. That's probably quite far in the future and I didn't want to future proof these tests too far in advance. 

It's probably not the most elegant function either, I just kind of wanted to have something running so we can test our work a bit easier o/

PR #2, opened 2018-03-15
Compass 2619 write a simple python generator v2

alenakhineika:


codegeneration/Python3Generator.js
```diff
@@ -2,66 +2,233 @@ const CodeGenerator = require('./CodeGenerator.js');
...
+  if (args.getChildCount() !== 3) {
+    return 'Error: Binary requires one or two arguments';
+  }
+
+  const childArgs = args.getChild(1);
+
+  if (childArgs.getChildCount() === 1) {
+    const data = this.visit(childArgs);
+
+    if (childArgs.type !== this.types.STRING) {
```
aherlihy: The binary type argument can be either a number or an enum. The way I got around this for java was to just define the enums in [this document](https://github.com/mongodb-js/bson-compilers/blob/master/codegeneration/JavaGenerator.js#L118-L126). The equivalent enums for python are defined here: http://api.mongodb.com/python/current/api/bson/binary.html

codegeneration/Python3Generator.js
```diff
@@ -2,66 +2,233 @@ const CodeGenerator = require('./CodeGenerator.js');
...
+  return `float(${double})`;
+};
+
+/**
+ * Visit Long Constructor
+ *
+ * @param {object} ctx
+ * @returns {string}
+ */
+Visitor.prototype.visitBSONLongConstructor = function(ctx) {
```
aherlihy: Because `Long` is a really complicated type with a lot of built-in functions, I thought it might be easiest if we execute it and then just extract the end result. [Example](https://github.com/mongodb-js/bson-compilers/blob/master/codegeneration/JavaGenerator.js#L175), that way we don't have to worry about supporting all the different `toString/Int/JSON`/etc functions directly.

codegeneration/Python3Generator.js
```diff
@@ -2,66 +2,248 @@ const CodeGenerator = require('./CodeGenerator.js');
...
+    const scope = this.visit(args.getChild(1).getChild(2));
+
+    if (args.getChild(1).getChild(2).type !== this.types.OBJECT) {
+      return 'Error: Code requires scope to be an object';
+    }
+
+    return `new Code(${code}, ${scope})`;
+  }
+
+  return `Code(${childArgs})`;
```
aherlihy: This should be `code`, not `childArgs`.

codegeneration/Python3Generator.js
```diff
@@ -2,66 +2,248 @@ const CodeGenerator = require('./CodeGenerator.js');
...
+
+  let longstr = '';
+
+  try {
+    longstr = this.executeJavascript(ctx.getText()).toString();
+  } catch (error) {
+    return error.message;
+  }
+
+  return `Int64(${longstr})`;
```
aherlihy: Does Int64 in Python accept a string or just a number? http://api.mongodb.com/python/current/api/bson/int64.html

alenakhineika: according to spec http://api.mongodb.com/python/current/api/bson/int64.html it is numeric value. and i have checked some examples like here https://stackoverflow.com/questions/30082047/does-long-or-int-in-pymongo-queries-have-to-match-the-number-type-in-mongodb they use numbers

aherlihy:
Changes included in https://github.com/mongodb-js/bson-compilers/pull/3

PR #4, opened 2018-03-16
Get Tests Passing

aherlihy:
[I'm going to update travis in a separate PR, so this will likely break master for a minute]
- Separate out language-constructors.json into two files, one for BSON constructors and one for built-in language types like literals, `Numbers`, `Date`, `RegExp`.
- Add a `description` field to each test and nest our tests by input/output language and type.
- Move helpers into test/helpers.js
- Move all code in the test files into tests
- Remove old, non-json tests
- Fix incorrect "expected" values in the json
- Add ability to skip unsupported fields so that we can start using the tests immediately. I thought it would be much nicer to be able to dynamically skip tests, and not keep anything in a gist, because I want to write tests as soon as I implement something and not wait until it's been implemented in all languages. Right now types that are not yet implemented are stored by language type in the `unsupported` object in helpers.js.
- Small fixes to the generators to match expected test results.
- Remove any code executed in index.js. Going forward we can still use it locally, but shouldn't commit our little tests since it leads to a lot of unnecessary changes.


lrlna:
THIS IS GR888888!! :sparkles:

PR #9, opened 2018-03-19
Add toPython and toJava

aherlihy:
Export language-specific functions so that package users don't have to construct their own generators. Will be much nicer to demo.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,3 @@
-/* eslint no-unused-vars: 0*/
 const antlr4 = require('antlr4');
 const ECMAScriptLexer = require('./lib/ECMAScriptLexer.js');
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -10,29 +9,20 @@ const JavaGenerator = require('./codegeneration/JavaGenerator.js');
...
-  return generator.visitExpressionSequence(tree);
+  return generator.start(tree);
 };
 
-const input = '[1, 2]';
-console.log(compileECMAScript(input, new JavaGenerator()));
-
-module.exports = compileECMAScript;
+module.exports = {
+  toJava: (input) => { return compileECMAScript(input, new JavaGenerator()); },
```
lrlna: yesssssssss!

PR #17, opened 2018-03-26
Compass 2626 expand csharp generator to bson constructors

alenakhineika:
Compass 2626 expand csharp generator to bson constructors


lrlna:
This looks great! I am going to merge :sparkles:

PR #18, opened 2018-03-27
make dates work in csharp

lrlna:
C# does the same thing as python -- needs date numbers as separate arguments. 

The only quirk is that the `now` constructor doesn't need the `new` word. not quite sure why, but tried it out in [C# playground](https://dotnetfiddle.net/) and it works as just `DateTime.Now`.


lrlna:
btw, if this is an okay PR, feel free to merge merge merge!

PR #19, opened 2018-03-27
regex output for c#

lrlna:
make regex get converted to c###### 

- [x] tests with appropriate results in c#
- [x] visitors to adjust the input to make the right output

I am currently using a few things to make sure this c# thing is working:
- [docs on regex options in c#](https://msdn.microsoft.com/en-us/library/system.text.regularexpressions.regexoptions(v=vs.110).aspx)
- [docs on regex itself](https://msdn.microsoft.com/en-us/library/system.text.regularexpressions.regex(v=vs.110).aspx)
- [regex playground for .net](http://regexstorm.net/tester)
- [playground for .net itself](https://dotnetfiddle.net/)

### DIFF
--- a/codegeneration/CSharpGenerator.js
+++ b/codegeneration/CSharpGenerator.js
@@ -1,5 +1,13 @@
 const CodeGenerator = require('./CodeGenerator.js');
 
+const CSHARP_REGEX_FLAGS = {
+  i: 'RegexOptions.IgnoreCase',
+  m: 'RegexOptions.Multiline',
+  u: '',
+  y: '',
+  g: ''
+};
+
 function Visitor() {
   CodeGenerator.call(this);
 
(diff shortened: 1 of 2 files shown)


lrlna:
we r g00d 2 go on this one

lrlna:
@aherlihy this PR is just for the built-in regex stuff, so it's a bit different than what we have with flags. fwik @alenakhineika worked on bson regex in #17. 

About the unicode stuff, there is no specific flag (the docs on this are [here](https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options)), but there is this thing about being able to do straight up in the regex:

>The language element \p{name}, which represents a Unicode category or named block, can't include embedded spaces in the \p{ portion of the element. If you do include a white space, the element throws an ArgumentException at run time. 

Now that I think about it, this might mean that instead of attaching a flag to the regex, I should be wrapping the input in the `\p{input}` thing?

aherlihy:
@lrlna I think you an attach a flag, since that's the simplest way of doing it. I think for now this is fine, we can always revisit. 

You can merge if you like, or you can just move this code directly into the emitRegex method in the csharp/generator.js file.

PR #20, opened 2018-03-27
[WIP] Add symbol table

aherlihy:
So this is a work in progress, but I figured I could go through what I've done so far and you could let me know if anything doesn't make sense. Apologies for the huge PR but it touches pretty much everything.

First thing, if you haven't already, it would be good to read up on symbol tables. The [wikipedia page](https://en.wikipedia.org/wiki/Symbol_table) is (surprisingly) good, but if you want something more in-depth, I highly recommend [this series of blog posts](https://eli.thegreenplace.net/2010/09/18/python-internals-symbol-tables-part-1). It's about how Python handles symbols internally, which is quite similar to what I'm doing but with a bit more complexity.

Basically, symbols are anything that is not a literal. So any variables, function calls, constructors, built-in methods, etc. Scope is just a data structure for keeping track of these symbols. Scope and namespaces are basically the same thing. 

I would recommend reading the code (especially the comments!) first, then this wall of text. Here I'm going to try to explain the reasoning behind the changes or various TODOs. Please comment with questions!

The changes can be summarized as:

1. `./grammars/ECMAScript.g4`: Remove all the BSON/JS constructors from the grammar. So now all of the BSON/JS constructors are going to be handled as if they were a function call. Why?
   - All type constructors are technically functions, so we should handle them like functions. This also means that if down the line we want to expand our language so that users can define their own classes, variables, or functions, it will be handled _exactly the same_ as the built-in types.
   - There are a *lot* of different built-in methods for each BSON class, and if we had a specific visitor function for each one our visitor code would get really long and complicated. There are special cases, namely when we want to execute the JS code, which are still handled by functions. These are now called `emit<Type>` instead of `visit<Node>` because it's not visiting a specific node anymore.
   - I eventually want to extract anything target-language-specific out of the visitor, except for the ones we really can't help like Date or RegExp.
2. `./codegeneration/SymbolTable.js` 
   - Right now the Symbol objects and the Scope objects are just wrappers, but eventually we'll need a data structure that can support fast lookups and nested scopes. I've put this off because it makes my head hurt and we don't care about it being slow at the moment (or honesty, ever, but we can put a little effort into to make it fast).
   - The primitive types are now stored in the `Types` table, and are Symbols instead of just enums. Eventually we can flesh out these types so that they have their own attributes, so we can support stuff like `"a string".length` or equiv.
   - *Right now I've just added the Java string template directly to the Symbols, but I'm going to extract it out so we can chose which template we want based off the target language. Not sure the best way to store those, TBD. Would love to hear if you have ideas!*
3. So for the test changes:
   - I've added test files for BSON object methods (i.e. `ObjectId().toString()`) and BSON utils (i.e. `ObjectId.fromTimestamp(<int>)`.
   -  Since the C# and Python generators still expect BSONConstructor nodes from the grammar, they will fail, so they're being skipped. '*' now skips all tests just to make life a little bit easier when developing.
   - It seems we completely forgot the `Int32` object, so I've added that.
4. We're now throwing errors in some places, instead of just returning strings, since when we have an error we want to stop everything and report it. This will be replaced entirely by #21.


codegeneration/SymbolTable.js
```diff
@@ -0,0 +1,412 @@
...
+    'MaxKey',
+    SYMBOL_TYPE.VAR, null, Types._object, // not sure this makes sense
+    new Scope({})
+  ),
+  BSONRegExp: new Symbol(
+    'BsonRegExp',
+    SYMBOL_TYPE.VAR, null, Types._object, // not sure this makes sense
+    new Scope({})
+  ),
+  Timestamp: new Symbol(
```
lrlna: i think what would be helpful is an explanation on how one of these structs is meant to work cause i am getting a bit confused with the whole nested situation of `Scope` --> `Symbol` --> `Scope` --> `Scope` 😅 😅 😅 i am really sorry if it's just me 😅 😅 😅

codegeneration/CodeGenerator.js
```diff
@@ -217,4 +266,76 @@ Visitor.prototype.executeJavascript = function(input) {
...
+};
+
+// /////////////
+// Emit type  //
+// /////////////
+/**
+ * @param {FuncCallExpressionContext} ctx
+ * @return {String}
+ */
+Visitor.prototype.emitType = function(ctx) {
```
lrlna: this is not meant to work as an actual event emitter, right? just a way to get the type of the current item we are visiting?

codegeneration/CodeGenerator.js
```diff
@@ -58,51 +56,112 @@ Visitor.prototype.visitChildren = function(ctx, options) {
...
-  ctx.type = this.types.INTEGER;
   return this.visitChildren(ctx);
 };
 
-Visitor.prototype.visitDecimalLiteral = function(ctx) {
-  ctx.type = this.types.DECIMAL;
-  return this.visitChildren(ctx);
+Visitor.prototype.visitFuncCallExpression = function(ctx) {
+  this.visit(ctx.singleExpression());
+  let lhsType = ctx.singleExpression().type;
```
lrlna: i was getting confused on this type thing a few days ago actually, so i am going to double check while i am here to double check if i understood this correctly.

1. visit the overarching expression -- string, number, boolean
2. assign a type to it while visiting
3. change whatever we got to change in that type
4. visit a more nested expression and apply changes based on the type we've assigned previously. (so a string would get a different treatment than a number kind of thing)

^^ Is that what we are doing?

lrlna: and then i guess once we use the symbol table we should be able to have more granular types? i guess the thing i don't get is how we actually assign the more granular types?

lrlna:
Errrr hi! So my "review" is just a bunch of questions that i had while reading these. they are not ordered since i started reading the code from the `SymbolTable` file. I am really sorry if the questions are kind of simple -- I've never dealt with Symbol Tables, or tuples or anything like that so I am just trying to wrap my head how the data (or arguments i guess) are going to flow through in this new setup o/ o/ o/ o/

i think i am going to draw this new system out next, and see if i understand it properly

codegeneration/SymbolTable.js
```diff
@@ -0,0 +1,412 @@
...
+const { doubleQuoteStringify } = require('./helpers');
+
+const SYMBOL_TYPE = { VAR: 0, CONSTRUCTOR: 1, FUNC: 2 };
+
+/**
+ * Symbols represent classes, variables, and functions.
+ *
+ * @param {String} id - identifier. TODO: for now, internals start with _
+ * @param {Number} callable - if it's a function, constructor, or variable.
+ * @param {Array} args - arguments if its callable. An array of tuples where
```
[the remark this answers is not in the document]
aherlihy: Tuples are basically just sets, and I didn't realize until after I wrote this comment that tuples don't really exist in javascript. So in this case we can just use an array of arrays. 

We need to store `args` so that we can type-check arguments to functions. Optional arguments are represented with null. I definitely don't think this is the neatest solution, but it lends itself to relatively straightforward type checking (in CodeGenerator:checkArguments). Can definitely change it.

So for something like the ObjectId constructor, which can take in either 0 arguments or 1 String/Number, args would look like: `[ [Types._string, Types._numerical, null] ]`. For something like the DBRef constructor, which takes in 2 arguments of type String and ObjectId, and an optional third String argument, the tuple would look like: `[ [Types._string], [BsonTypes.ObjectId], [Types._string, null] ]`.

codegeneration/SymbolTable.js
```diff
@@ -0,0 +1,412 @@
...
+ * each tuple has each possible type for the argument at that index.
+ * @param {Symbol} type - the type the symbol returns. Could be a Symbol or
+ * 0 if it's a primitive type.
+ * @param {Scope} attrs - the attributes of the returned type. TODO: do we want to strictly check all objs or just BSON/Built-in.
+ * @param {Function} template - the string template for this type. This is the first
```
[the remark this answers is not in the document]
aherlihy: Date is special cased because we have to execute JS code to extract the date that we want. But basically, yes. The template is the translation of the symbol into the target language, so it will change depending on what language we're compiling to. The next major TODO (beyond cleaning up/etc) is to allow us to swap out the template argument in the symbol table depending on target language. That way, eventually, adding a new target language could just be filling in the templates.

codegeneration/SymbolTable.js
```diff
@@ -0,0 +1,412 @@
...
+    callable: callable,
+    args: args,
+    type: type,
+    attr: attrs,
+    template: template
+  };
+}
+
+/**
+ * Scope represents both namespaces and variable scope. Eventually the
```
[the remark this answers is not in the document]
aherlihy: Right now we only have a concept of global scope, i.e. everything is either defined or undefined. Pushing and popping scopes will come whenever we implement control flow (like loops/etc) and function definitions. When we add that, we'll need to change the Scope datastructure from being a wrapper to something that can do scoped lookup.

The caveat is that you can sort of think of namespaces as if they were scopes. i.e. if you have a `fromString` function on it's own, it's undefined, but if you have `ObjectId.fromString` then it's in scope. That's why there are nested scopes within symbols. I'm really not sure if this is the "right" way to do it, this is just my best guess. It's been a long time since I did anything with compilers so I could be missing something.

Generally, when we have local scope it will be set by the visitor when it enters a scoped block, i.e. the new scope gets pushed onto the symbol table. For example, if you enter something like a function definition block, the arguments get pushed onto the scope. If you enter an if statement, the scope that gets pushed on is empty. When you leave the block, the scope (along with any symbols that had been added within that block) gets popped.

codegeneration/SymbolTable.js
```diff
@@ -0,0 +1,412 @@
+/* eslint key-spacing:0, no-multi-spaces:0 */
+const { doubleQuoteStringify } = require('./helpers');
+
+const SYMBOL_TYPE = { VAR: 0, CONSTRUCTOR: 1, FUNC: 2 };
```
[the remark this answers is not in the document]
aherlihy: Usually, symbol tables only care if a symbol callable or not, i.e. if you have a symbol `x`, if `x()` is an error or not. However, for us, we also care if it's an object constructor because at least for Java, if it's a constructor, we need to tack on a `new` before the call. What this enum is determining is 1. if we can call it, and 2. if we need to add new before it.

codegeneration/SymbolTable.js
aherlihy: It's not just you, it's really confusing. I kinda explained it in an earlier comment, but the idea is:

Symbol is an identifier. Scope is just a set of identifiers.

So we have our "Global Scope", meaning the set of symbols that are always accessible. Symbols have namespaces associated with them. Namespaces are a set of symbols. Thus namespaces are a scope as well.

Maybe another useful way to think about it would be when a visitor visits a identifier, it only has the string. In order to compile it correctly, it needs to be able to look up:
  1. The type of the symbol, i.e. is it a function? a constructor? a variable? (`callable`)
  2. If it's a function, what is the function signature, i.e. what arguments does it accept? (`args`). If it's a variable, args is null.
  3. The other half of a function signature is the return type of the function.  i.e. is it a string? number? object instance? We also need to know the type if it's a variable, so that we can correctly type-check. (`type`).
  4. What attributes does it have, i.e. what is in it's namespace? We need this for when we visit `memberDotExpression`s, i.e. `field.member`. When visiting the expression we need to check if member is in field's scope or not. (`attr`).
  5. What is this symbol's name in our target language? This will only apply to js-types and bson-types, since anything user-defined will be the same no matter the target language. (`template`).

Was that helpful? Something confusing is that we're using Symbols to represent both types *and* stored variables. Since they are kind of the same thing. I thought a lot about separating out types and symbols into different classes/tables, but ultimately decided to treat them all as symbols since we need to store mostly the same information. A lot of the reason that this is complicated stems from the fact that `ObjectId()` and `ObjectId` are exactly the same symbol, but have totally different set of attributes.

codegeneration/CodeGenerator.js
aherlihy: No, it's called emit because that's what I called it in the last compiler that I wrote....and obviously can't remember why I picked emit. Probably because it spits out the type in the target language. Can be renamed, forgot that emit is a bigger thing in javascript :)

codegeneration/CodeGenerator.js
aherlihy: I'm not sure I understand the question, could you be more specific about what you're double checking? You mean what the visitFuncCallExpression method does?

codegeneration/SymbolTable.js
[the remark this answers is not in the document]
lrlna: > when we have local scope it will be set by the visitor when it enters a scoped block, i.e. the new scope gets pushed onto the symbol table

ok that makes a *lot* more sense. it seems to act like a fancy `stack` that has a lot more meaning tied to it. I can totally reason about something like that!

codegeneration/SymbolTable.js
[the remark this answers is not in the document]
lrlna: ahhhh okay this is just a way for us to figure out with what kind of item we are dealing with! got it!

codegeneration/CodeGenerator.js
lrlna: haha cool yea no worries! i think now that i understand other stuff, this is more clear too

codegeneration/CodeGenerator.js
lrlna: i was more inquiring about the `type` stuff, but now that i read through your explanation on the way we are meant to use the symbol table this question is irrelevant :)

codegeneration/SymbolTable.js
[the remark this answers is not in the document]
lrlna: ahhhhhh so the `args` here  map to the arguments the constructor takes for the item we are visiting. they are not `args` for the literal `new Symbol()`. They are just meant to describe what we want to work with when "entering" a particular item with our visitor. this makes more sense then!

codegeneration/SymbolTable.js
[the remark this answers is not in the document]
lrlna: cool! yea! that's pretty clear now then !

codegeneration/SymbolTable.js
lrlna: I think I get it a little bit better, yea! thank you ^___^ i've got it structured like this, which is essentially what you have in code version here but laid down horizontally hahaha

codegeneration/SymbolTable.js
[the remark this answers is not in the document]
lrlna: ok i drew it out and it makes more sense now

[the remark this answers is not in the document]
aherlihy: Beautiful!!!!

codegeneration/SymbolTable.js
aherlihy: Yeah, these are awesome! We should put them in the README when we're all done with this :)

alenakhineika:
Anna, you have done amazing job! i cannot recommend anything on top of it, i can understand the concept, but all questions will appear during working with it :) and thanks a lot for documentation section and comment i the code, they are really helpful!

lrlna:
WOO for the merge!! :sparkles:

PR #21, opened 2018-03-28
Errors handling

alenakhineika:
Errors handling


aherlihy:
This looks great! Do you want to add tests for it as part of this pull request? We can keep the tests in JSON format, just write a different `test.helpers.js:runTest` function that expects an error to be thrown and checks the message.

alenakhineika:
@aherlihy @durran i have updated PR with tests support. You can see one example of using it in Python3Generator for 'code' constructor. Since antlr returns any error as valid string and it makes difficulties to parse it, i have changed this behaviour and am throwing an error, which can be caught. Pls let me know your opinion about it. If you like this approach, I will update rest of errors this way. Or correct me and i will fix it. Thx in advance!

durran:
LGTM. I think throwing the errors is more expected behaviour in my opinion.

codegeneration/AppError.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Custom Application Error
+ *
+ * @returns {object}
+ */
+class AppError extends Error {
```
aherlihy: I think we should probably give this a more specific name than AppError, since it'll be imported into whatever packages end up using the compilers

index.js
```diff
@@ -6,23 +6,41 @@ const Python3Generator = require('./codegeneration/Python3Generator.js');
...
  *
  * @param {String} input
  * @param {CodeGenerator} generator
  * @returns {String}
  */
-const compileECMAScript = function(input, generator) {
+const compileECMAScript = (
+  input,
+  generator
+) => new Promise((resolve, reject) => {
```
aherlihy: Would you mind explaining to me why we need this to be a promise? Thanks!

codegeneration/Python3Generator.js
```diff
@@ -1,684 +1,691 @@
...
-  return `${key}: ${value}`;
-};
+    if (argList.length === 2) {
+      /* NOTE: we have to visit the subtree first before type checking or type may
+      not be set. We might have to just suck it up and do two passes, but maybe
+      we can avoid it for now. */
+      const scope = this.visit(argList[1]);
 
+      if (argList[1].type !== this.types.OBJECT) {
+        return 'Error: Code requires scope to be an object';
```
aherlihy: There are still some instances where strings are still being returned instead of exceptions.

codegeneration/Python3Generator.js
```diff
@@ -1,684 +1,691 @@
...
+   * @returns {string}
+   */
+  visitBSONCodeConstructor(ctx) {
+    const argumentList = ctx.arguments().argumentList();
+
+    if (
+      argumentList === null ||
+      (argumentList.getChildCount() !== 1 && argumentList.getChildCount() !== 3)
+    ) {
+      throw new AppError({
```
aherlihy: This is great! I was going to ask if you wanted to subclass the errors to be more specific (that's how it would be done in Python, I'm not sure if that's normal for JS or not) or just set the codes directly. Not sure what's more canonical

codegeneration/AppError.js
alenakhineika: What about BsonCompilersAppError? Or pls don't hesitate to advice me any better variant!

index.js
durran: Can't Compass just `catch (e)` and get any error that was thrown and then print `e.message`? That would be my expectation. By using a promise here I am expecting async I/O to be occurring which isn't the case here - the code is actually synchronous but is faking being async, which is a bit odd.

alenakhineika: yes it can be done like this. i will remove promise then.

codegeneration/AppError.js
aherlihy: I think we can make it more generic - something like CompilerError since the error could be from any code not just BSON-related. SyntaxError can inherit from CompilerError, and we can make as many other subclasses as we like.

codegeneration/Python3Generator.js
aherlihy: I think that sounds great! These are the types of errors I was thinking about, does this seem like a good start? We can always add more.

1. type errors
2. argument count mismatch errors
3. undefined symbol
4. generic error, for the stuff that doesn't fit in anywhere else. So for example if we catch an exception while executing the JS, we can create the "other" error with the message that was set by the JS exception that we caught.
6. not-yet-implemented error: so if the user tries to use some kind of feature that we don't support, like lambdas for example, we can throw this.

package.json
```diff
@@ -32,6 +32,7 @@
     "antlr4": "^4.7.1",
     "bson": "^2.0.4",
     "chai": "^4.1.2",
+    "chai-as-promised": "^7.1.1",
```
aherlihy: Don't think we need this anymore without the promise?

durran:
Great job!

helper/error.js
```diff
@@ -0,0 +1,36 @@
...
+
+        this.code = code;
+        this.message = message;
+
+        if (typeof payload !== 'undefined') {
+          this.message = payload.message || message;
+          this.payload = payload;
+        }
+
+        Error.captureStackTrace(this, errors[name]);
```
lrlna: i am late to the game, but we might want to some time in the future have a `try/catch` around this cause `Error.captureStackTrace()` will not be available in the browser (it's node's error property).

PR #22, opened 2018-04-09
Errors handling and refactoring for using classes instead of prototypes

alenakhineika:
Errors handling and refactoring for using classes instead of prototypes


lrlna:
I can't find how the error handling got changed in the diff. Was there a change, or you just added extra error handling things?

alenakhineika:
@lrlna Instead of returning just a string error:

´return 'Error: BSONRegExp requires pattern to be a string';´

now we are throwing an error:

´throw new SemanticArgumentCountMismatchError({message: 'Number requires one argument'});´

The error handling by itself was implemented in previous PR:

https://github.com/mongodb-js/bson-compilers/pull/21

In current PR i finished replacing errors for all code generators.

lrlna:
@alenakhineika ah okay cool! thank you for clarification. I am almost done digesting through #21, i just wasn't sure if this was adding something new, and couldn't find it in the diff ^__^

codegeneration/CodeGenerator.js
```diff
@@ -13,376 +13,376 @@ const {
...
+    }
 
-  const expectedArgs = lhsType.args;
-  let rhs = this.checkArguments(expectedArgs, ctx.arguments().argumentList());
+    const expectedArgs = lhsType.args;
+    let rhs = this.checkArguments(expectedArgs, ctx.arguments().argumentList());
 
-  ctx.type = lhsType.type;
-  if (!lhsType.callable) {
-    throw new SemanticArgumentCountMismatchError({
```
aherlihy: In this case, the error should be TypeError: `${lhs} is not a function (${lhsType.id})`

aherlihy: Sorry this should be on the new code, not the deleted one, so line 159

codegeneration/CodeGenerator.js
```diff
@@ -13,376 +13,376 @@ const {
...
-      message: `symbol "${name}" is undefined`
-    });
-  }
-  if (ctx.type.template) {
-    return ctx.type.template();
+  visitIdentifierExpression(ctx) {
+    const name = this.visitChildren(ctx);
+    ctx.type = this.Symbols[name];
+    if (ctx.type === undefined) {
+      throw new SemanticArgumentCountMismatchError({
```
aherlihy: This should be a ReferenceError: `${name} is not defined`

codegeneration/CodeGenerator.js
```diff
@@ -13,376 +13,376 @@ const {
...
-      if (typeof type === 'string') {
-        type = this.Types[type];
+    let type = ctx.singleExpression().type;
+    if (typeof type === 'string') {
+      type = this.Types[type];
+    }
+    while (type !== null) {
+      if (!(type.attr.hasOwnProperty(rhs))) {
+        if (type.id in this.BsonSymbols && this.BsonSymbols.type.id !== null) {
+          throw new SemanticArgumentCountMismatchError({
```
aherlihy: This should be an AttributeError: `${rhs} not an attribute of ${type.id}`

codegeneration/CodeGenerator.js
```diff
@@ -13,376 +13,376 @@ const {
...
+          args[i].type === this.Types._octal)) {
+        continue;
+      }
+      if (expected[i].indexOf(args[i].type) === -1 && expected[i].indexOf(args[i].type.id) === -1) {
+        const message = `expected types ${expected[i].map((e) => {
+          return e.id ? e.id : e;
+        })} but got type ${args[i].type.id} for argument at index ${i}`;
 
-      throw new SemanticArgumentCountMismatchError({message});
+        throw new SemanticArgumentCountMismatchError({message});
```
aherlihy: This should be a TypeError

codegeneration/CodeGenerator.js
```diff
@@ -13,376 +13,376 @@ const {
...
-    argStr.push(this.visit(args[i]));
-    if (expected[i].indexOf(this.Types._numeric) !== -1 && (
-        args[i].type === this.Types._integer ||
-        args[i].type === this.Types._decimal ||
-        args[i].type === this.Types._hex ||
-        args[i].type === this.Types._octal)) {
-      continue;
+    const args = argumentList.singleExpression();
+    if (args.length > expected.length) {
+      throw new SemanticArgumentCountMismatchError({message: 'too many arguments'});
```
aherlihy: We should add a more specific error message here, so "arguments mismatch: expected x and got y"

PR #32, opened 2018-04-18
remove path-resolve

lrlna:
`path.resolve()` does not work properly in a browser environment. This PR changes all the `path.resolve()` require statements to direct links to files.

If this proves too messy in the future, we can possibly consider splitting off error handling (all the `path.resolve()` happens for error handling portion of this package) into its own helper module we can publish / maintain / require separately.


lrlna:
I am going to merge this PR without waiting for a review since I am trying to trouble shoot something in a requiring it in an electron env. Buttttt we can *always* revert this publish if anyone has problems with it or has other suggestions with dealing with this in the browser!!

aherlihy:
The reason we were using path.resolve was so that it would work on any operating system...but if we're going to use browserify to get rid of any filesystem access at all, then it won't matter, right? And we don't develop on windows, so there's not too much point in doing anything to make that work.

PR #33, opened 2018-04-18
COMPASS-2754 Figure out how to test non-symmetric translations

aherlihy:
Previously our test suite expected all translations to be symmetric, but that's not always possible (especially as our generated code gets more idiomatic). 

For example, Code in javascript can accept a function or a string, but in shell it only accepts a string.

Javascript --> Shell:

```
Code(function(x){console.log(x)};) --> Code("function(x){console.log(x)};")
```
Shell --> Javascript 
```
Code("function(x){console.log(x)};") --> Code("function(x){console.log(x)};")
```

So now we have a folder for each output language, and in that folder the JSON files will only be tested from input language --> target language.


aherlihy:
Dup of https://github.com/mongodb-js/bson-compilers/pull/34

PR #38, opened 2018-04-20
Generate numeric literals in javascript as Long in java

aherlihy:
Default numeric type in java is int32, so regular numbers in javascript will get translated as long.

Do this everywhere, except in special cases where we want to keep it an int (for example, `new Long(1)` should not translate to `new java.lang.Long(new java.lang.Long(1))`.


aherlihy:
Replaced by https://github.com/mongodb-js/bson-compilers/pull/39

PR #41, opened 2018-04-24
Add shell bson constructors as input language

aherlihy:
Instead of listing everything I've done here, I'm going to try something new and make comments on the code itself. Let me know if it's easier to follow.

This PR adds support for the shell input language. Includes the bson and js constructors, the methods and utils are TODO.


codegeneration/java/Generator.js
```diff
@@ -85,18 +85,24 @@ module.exports = (superClass) => class ExtendedVisitor extends superClass {
    * @return {String}
    */
   emitDate(ctx) {
+    let toStr = '';
```
aherlihy: Fixes the behavior for calling `Date()` in javascript and shell. I didn't realize that `Date(x)` actually returns a string, while `new Date(x)` returns the date object. The `emitDate` method has been updated in java/Generator to support this, and new tests have been added that need to be filled out for python/csharp.

codegeneration/java/Generator.js
```diff
@@ -85,18 +85,24 @@ module.exports = (superClass) => class ExtendedVisitor extends superClass {
...
+      return `new java.util.Date()${toStr}`;
     }
     let epoch;
     try {
       epoch = this.executeJavascript(ctx.getText()).getTime();
     } catch (error) {
       throw new SemanticGenericError({message: error.message});
     }
-    return `new java.util.Date(${epoch})`;
+    return `new java.util.Date(new java.lang.Long(${epoch}))${toStr}`;
```
aherlihy: Most of the epoch numbers are too large for int32, so now constructing from Long so that it will compile in java.

codegeneration/javascript/Visitor.js
```diff
@@ -329,18 +330,6 @@ class Visitor extends ECMAScriptVisitor {
       Code: function(c, s) {
         return new bson.Code(c, s);
       },
-      NumberDecimal: function(s) {
```
aherlihy: These are shell types, and were left over from the old compass query bar implementation. Now we want to *only* support javascript here, and *only* support shell in shell/Visitor.js.

codegeneration/javascript/Visitor.js
```diff
@@ -409,7 +398,7 @@ class Visitor extends ECMAScriptVisitor {
   checkArguments(expected, argumentList) {
     const argStr = [];
     if (!argumentList) {
-      if (expected.length === 0) {
+      if (expected.length === 0 || expected[0].indexOf(null) !== -1) {
```
aherlihy: Bugfix.

codegeneration/shell/Visitor.js
```diff
@@ -16,6 +16,25 @@ class Visitor extends JavascriptVisitor {
     this.new = '';
   }
 
+  /**
+   * New in the shell is the same as calling without arguments.
+   * @param {NewExpressionContext} ctx
+   * @return {String}
+   */
+  visitNewExpression(ctx) {
+    ctx.singleExpression().wasNew = true;
```
aherlihy: In shell, `new Funcname` is the same as calling `Funcname()`.

What this is doing is transforming the tree so that if we have a new call on a non-function node, then we force it to be visited as if it was a function call with no arguments.

codegeneration/shell/Visitor.js
```diff
@@ -30,12 +49,18 @@ class Visitor extends JavascriptVisitor {
         return new bson.Code(c, s);
       },
       NumberDecimal: function(s) {
-        return bson.Decimal128.fromString(s);
+        if (s === undefined) {
```
aherlihy: adding support no arguments meaning 0.

symbols/java/templates.yaml
```diff
@@ -17,7 +17,7 @@ Templates:
     IntegerTypeTemplate: &IntegerTypeTemplate null
     DecimalTypeTemplate: &DecimalTypeTemplate !!js/function >
         (literal) => {
-            return `new java.lang.Float(${literal})`;
```
aherlihy: Bugfix.

symbols/java/templates.yaml
```diff
@@ -339,8 +349,14 @@ Templates:
             return `${lhs}.toString`;
         }
     # Symbol Templates
-    CodeSymbolTemplate: &CodeSymbolTemplate null # Has emit method
-    CodeSymbolArgsTemplate: &CodeSymbolArgsTemplate null
+    CodeSymbolTemplate: &CodeSymbolTemplate  !!js/function >
```
aherlihy: Since shell doesn't accept functions as code arguments, we don't need an emit method for the shell input language. We can just use templates.

test/json/success/javascript/js-constructors.json
```diff
@@ -158,15 +158,55 @@
...
         "javascript": "new Date(1995, 11, 17, 3, 24, 0)",
         "python": "datetime.datetime(1995, 12, 17, 3, 24, 0, tzinfo=datetime.timezone.utc)",
-        "java": "new java.util.Date(819170640000)",
+        "java": "new java.util.Date(new java.lang.Long(819170640000))",
         "csharp": "new DateTime(1995, 12, 17, 3, 24, 0)"
+      },
+      {
+        "description": "Date string with ISO String",
+        "javascript": "Date('December 17, 1995 03:24:00Z')",
+        "python": "",
```
aherlihy: This needs to be implemented for python/csharp

codegeneration/shell/Visitor.js
lrlna: i think i am confused with this one a bit. wouldn't it be visited anyway because there is a function call already, i.e. `new Morph()`? or are you saying there are sometimes instances where we just have `new Morph` (without the `(` `)`)?

test/json/success/javascript/js-constructors.json
lrlna: 👍

codegeneration/java/Generator.js
lrlna: whooaaaa TIL! I thought it was an object always for some reason, but then again I only ever do `new Date()` :O

lrlna:
Ok, I know this probably took you *so* much longer to do than putting it in a body of a PR, but it's *sooo* much nicer to read and understand when you've commented on individual pieces of code 👏 👏 👏

codegeneration/shell/Visitor.js
aherlihy: Exactly, in the shell (and in javascript, TIL) `new Morph` is totally valid and equivalent to `new Morph()`. idk if people use it....but it works? I think it's used with like `new MinKey` instead of writing out `new MinKey()`.

aherlihy: @lrlna I've moved this into the javascript visitor because this behavior also works in regular javascript.

PR #42, opened 2018-04-25
c# symbols for bson constructors

lrlna:
c# bson constructors now use the `template`.

I'll also comment on the code individually to explain what's going on, since that seems to be easier to read!


test/json/success/javascript/bson-constructors.json
```diff
@@ -115,15 +115,15 @@
         "javascript": "new DBRef('coll', new ObjectId())",
         "python": "DBRef('coll', ObjectId())",
         "java": "new DBRef(\"coll\", new ObjectId())",
-        "csharp": "DBRef(\"coll\", new BsonObjectId())",
+        "csharp": "new MongoDBRef(\"coll\", new BsonObjectId())",
```
lrlna: `DBRef` doesn't exist in c# apparently! Gonna use this `MongoDBRef` as per [the docs](https://mongodb.github.io/mongo-csharp-driver/2.4/apidocs/html/M_MongoDB_Driver_MongoDBRef__ctor.htm)

codegeneration/csharp/Generator.js
```diff
@@ -1,136 +1,209 @@
...
-    return this.doubleQuoteStringify(this.visitChildren(ctx));
-  }
-
-  // there is no undefined in c#
-  visitUndefinedLiteral(ctx) {
-    ctx.type = this.Types._undefined;
+  constructor() {
+    super();
+    this.new = 'new ';
+    this.regexFlags = {
```
lrlna: i am going to keep these here for now, but there isn't a `emit` method for them just yet!

codegeneration/csharp/Generator.js
```diff
@@ -1,136 +1,209 @@
...
+   * @param {BSONInt32Object} ctx
+   *
+   * @returns {string} - Int32.Parse("value") OR Convert.ToInt32(value)
+   */
+  emitInt32(ctx) {
+    ctx.type = this.Types.Int32;
+    const args = ctx.arguments().argumentList().singleExpression();
+    const expr = args[0].getText();
+    if (expr.indexOf('\'') >= 0 || expr.indexOf('"') >= 0) {
+      return `Int32.Parse(${doubleQuoteStringify(expr.toString())})`;
```
lrlna: c# handles `string` and `int` params differently, so i am checking what's what before returning. alternatively i can possibly just execute js and just return one thing i suppose?

codegeneration/csharp/Generator.js
```diff
@@ -1,136 +1,209 @@
...
+   * BSON MinKey Constructor
+   * needs to be in emit, since does not need a 'new' keyword
+   *
+   * @param {BSONMinKeyObject} ctx
+   *
+   * @returns {string} - BsonMinKey.Value
+   */
+  emitMinKey(ctx) {
+    ctx.type = this.Types.MinKey;
+    return 'BsonMinKey.Value';
```
lrlna: there are a few methods that don't need a `new` keyword, like this one. so i am sticking it into the `Generator` instead of `template`.

codegeneration/csharp/Generator.js
```diff
@@ -178,284 +250,97 @@ module.exports = (superclass) => class ExtendedVisitor extends superclass {
...
+    ctx.type = this.Types.Decimal128;
 
+    let decimal;
     try {
-      hexstr = this.executeJavascript(ctx.getText()).toHexString();
+      decimal = this.executeJavascript(`new ${ctx.getText()}`);
     } catch (error) {
       throw new SemanticGenericError({message: error.message});
     }
+    const value = parseInt(decimal.toString(), 10);
```
lrlna: `Decimal128` only takes an `int32`, and i think this will get me an `int32`. i am getting super confused with all the number types tho hahah

symbols/csharp/templates.yaml
```diff
@@ -1,111 +1,401 @@
...
+        }
     NumericTypeTemplate: &NumericTypeTemplate null
-    ArrayTypeTemplate: &ArrayTypeTemplate null
-    NullTypeTemplate: &NullTypeTemplate null
-    UndefinedTypeTemplate: &UndefinedTypeTemplate null
-    ObjectTypeTemplate: &ObjectTypeTemplate null
-    ObjectTypeArgsTemplate: &ObjectTypeArgsTemplate null
+    ArrayTypeTemplate: &ArrayTypeTemplate !!js/function >
+        (literal) => {
+          if (literal === '') return 'new BsonArray()'
```
lrlna: there is a different output for when there are no arguments and there are some arguments to both `BsonArray` and `BsonDocument`, so i am doing a lil if/else here

symbols/csharp/templates.yaml
```diff
@@ -1,111 +1,401 @@
...
+          return 'BsonNull.Value';
+        }
+    UndefinedTypeTemplate: &UndefinedTypeTemplate !!js/function >
+        () => {
+          return 'BsonUndefined.Value';
+        }
+    ObjectTypeTemplate: &ObjectTypeTemplate !!js/function >
+        (literal) => {
+          if (literal === '') return `new BsonDocument()`;
+          return `new BsonDocument${literal}`;
```
lrlna: same as above ^^ need to do a lil if/else based on whether or not there are any args.

symbols/csharp/templates.yaml
```diff
@@ -1,111 +1,401 @@
...
+            if (
+                (pair[0].charAt(0) === '\'' && pair[0].charAt(pair[0].length - 1) === '\'') ||
+                (pair[0].charAt(0) === '"' && pair[0].charAt(pair[0].length - 1) === '"')) {
+                    newStr = pair[0].substr(1, pair[0].length - 2);
+                    console.log('new str in if', newStr)
+            }
+            newStr = `"${newStr.replace(/\\([\s\S])|(")/g, '\\$1$2')}"`;
+
+            if (str !== '') {
+              return `{ ${str.replace(/\((.*)\)/, "{ $1 }")}, { ${newStr}, ${pair[1]} } }`;
```
lrlna: if there are multiple objects being added, we need to have `{}`, if there is only one object being added, it's just the `()`

symbols/csharp/templates.yaml
```diff
@@ -1,111 +1,401 @@
...
-    TimestampGreaterThanOrEqualArgsTemplate: &TimestampGreaterThanOrEqualArgsTemplate null
-    TimestampLessThanTemplate: &TimestampLessThanTemplate null
-    TimestampLessThanArgsTemplate: &TimestampLessThanArgsTemplate null
-    TimestampLessThanOrEqualTemplate: &TimestampLessThanOrEqualTemplate null
-    TimestampLessThanOrEqualArgsTemplate: &TimestampLessThanOrEqualArgsTemplate null
-    SymbolValueOfTemplate: &SymbolValueOfTemplate null
-    SymbolInspectTemplate: &SymbolInspectTemplate null
-    SymbolToJSONTemplate: &SymbolToJSONTemplate null
-    SymbolToStringTemplate: &SymbolToStringTemplate null
+    ObjectIdToJSONTemplate: &ObjectIdToJSONTemplate !!js/function >
```
lrlna: all of these `BSON Object Methods` are very different in documentation between c# and java, so they might not be all correct at the moment. Probably should ask someone who works with the C# driver what the equivalent stuff is.

codegeneration/csharp/Generator.js
aherlihy: I think this is better, anytime we execute JS we're just going to make more work for us later when we're accepting a language that we can't execute :)

codegeneration/csharp/Generator.js
```diff
@@ -1,136 +1,209 @@
...
+   */
+  emitLong(ctx) {
+    ctx.type = this.Types.Long;
+    let longstr;
+    try {
+      longstr = this.executeJavascript(ctx.getText()).toString();
     } catch (error) {
       throw new SemanticGenericError({message: error.message});
     }
+    return `new BsonInt64(Convert.ToInt32(${longstr}))`;
```
aherlihy: C# is a weird language....

codegeneration/csharp/Generator.js
```diff
@@ -178,284 +250,97 @@ module.exports = (superclass) => class ExtendedVisitor extends superclass {
...
-    }
-
-    const argumentListExpression = argumentList.singleExpression();
-    const bytes = this.doubleQuoteStringify(binobj.toString());
-
-    if (argumentListExpression.length === 1) {
-      return `new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(${bytes}))`;
-    }
+  // c# does not have octal numbers, so we need to convert it to reg integer
+  // TODO: not sure if we should still set the type to OCTAL or INTEGER
```
aherlihy: I think that we can set it to Integer and not octal, since it doesn't exist anywhere else.

codegeneration/csharp/Generator.js
```diff
@@ -178,284 +250,97 @@ module.exports = (superclass) => class ExtendedVisitor extends superclass {
...
-    const argumentList = ctx.arguments().argumentList();
+  /*  ************** built-in js identifiers **************** */
 
-    if (argumentList === null || argumentList.getChildCount() !== 1) {
+  // adjust the Number constructor;
+  // returns new int(num)
+  visitNumberConstructorExpression(ctx) {
+    const argList = ctx.arguments().argumentList();
+
+    if (!argList || argList.singleExpression().length !== 1) {
```
aherlihy: I think you can use `checkArguments` here instead of checking them by hand, right?

symbols/csharp/templates.yaml
aherlihy: Yes! You should talk to Craig, I will send a message now.

codegeneration/csharp/Generator.js
```diff
@@ -1,136 +1,209 @@
...
-
+  /**
+   * BSON Binary Constructor
+   * needs to execute JS to get value first
+   *
+   * @param {BSONBinaryObject} ctx
+   *
+   * @returns {string} - new BsonBinaryData()
+   */
+  emitBinary(ctx) {
```
alenakhineika: According to current master it should be emitBinaryFromJS

codegeneration/csharp/Generator.js
alenakhineika: do we need emit method in this case? maybe we can use templates instead?

codegeneration/csharp/Generator.js
lrlna: @alenakhineika yea you're right! I worked off an old master version; need to rebase.

codegeneration/csharp/Generator.js
lrlna: yep! that's what i did with js-constructor stuff! I'll just PR all of that work after we merge this so it's not too many PRs open at the same time.

codegeneration/csharp/Generator.js
lrlna: @alenakhineika i wouldn't mind doing in templates, but I don't know how to omit the `new` keyword there. these two outputs -- `Int32.Parse()` and `ConvertToInt32()` -- don't need the `new` keyword unlike most other function calls with c#.

codegeneration/csharp/Generator.js
lrlna: all right, i rebased and it's now up-to-date with master's symbol table ids

lrlna:
errrr i don't know why all of these `lib` files are coming through as well. We ignore those, right?

.gitignore
```diff
@@ -4,6 +4,7 @@ npm-debug.log
 node_modules/
 lib-cov
 coverage
+lib/
```
aherlihy: This shouldn't be here

PR #44, opened 2018-04-27
c# js-constructors

lrlna:
wait to merge this PR after #42, since it relies on some changes in there. This will be a much smaller PR once the previous one is closed.

similarly, I'll comment out on the pieces of code explaining what's happening.


codegeneration/csharp/Generator.js
```diff
@@ -178,429 +185,164 @@ module.exports = (superclass) => class ExtendedVisitor extends superclass {
...
-   * @returns {string}
+   * @returns {string} - DateTime.Now
    */
-  visitNullLiteral(ctx) {
-    ctx.type = this.Types._null;
 
-    return 'BsonNull.Value';
+  emitnow(ctx) {
+    ctx.type = this.Types.Now;
+    return 'DateTime.Now';
```
lrlna: when you're constructing a `Date.Now()` object in c#, it not a new instance of the `Date` object, so we need to handle it in an `emit` method. similarly, `Now` is not a callable method, but just an attribute

codegeneration/csharp/Generator.js
```diff
@@ -178,429 +185,164 @@ module.exports = (superclass) => class ExtendedVisitor extends superclass {
...
-        message: 'Decimal128 requires one argument'
-      });
-    }
+    // we need to pipe ( "|" ) flags in csharp if there is more than one of them
+    const csharpflags = flags.replace(/[imuyg]/g, (m) => {
+      if (m === flags[flags.length - 1]) {
+        return this.regexFlags[m];
+      }
+      if (this.regexFlags[m] !== '' && flags.length > 1) {
+        return this.regexFlags[m] + ' | ';
```
lrlna: c# is weird, and instead of individual flags, we pipe together `RegexOptions` and their applicable values. so the `RegexOptions` object has all the `Multiline`, `Unicode`, etc methods.

test/json/success/javascript/js-constructors.json
```diff
@@ -66,28 +66,28 @@
...
-        "csharp": "1234567"
+        "csharp": "342391"
       },
       {
         "description": "Octal 00-prefix number",
         "javascript": "001234567",
         "python": "0o1234567",
         "java": "01234567",
-        "csharp": "1234567"
+        "csharp": "342391"
```
lrlna: I think i've mentioned this previously, but there are no octals in c#, so i am just straight up converting this into a number in the templates.

symbols/csharp/templates.yaml
```diff
@@ -1,147 +1,434 @@
...
     RegexTypeTemplate: &RegexTypeTemplate null
     BoolTypeTemplate: &BoolTypeTemplate null
     IntegerTypeTemplate: &IntegerTypeTemplate null
     DecimalTypeTemplate: &DecimalTypeTemplate null
     LongBasicTypeTemplate: &LongBasicTypeTemplate null
     HexTypeTemplate: &HexTypeTemplate null
-    OctalTypeTemplate: &OctalTypeTemplate null
+    OctalTypeTemplate: &OctalTypeTemplate !!js/function >
+        (literal) => {
+          literal = literal.replace(/[oO]+/g, '0')
```
lrlna: I have to hack around the issue of us always dealing with strings when arguments come in. This specific case is to account for the fact that when you run `parseInt("0o1234567", 10)` in JS, you get a different output (output = 0) then when you run `parseInt(0o1234567, 10)` (output=342391). To make sure I get the same number and the number I want, I convert every `O` into a `0` and then run the `parseInt` function. This works because substituting one for the other returns the number that we want.

symbols/csharp/templates.yaml
```diff
@@ -1,147 +1,434 @@
...
+          return parseInt(literal, 8)
```
lrlna: base 8 here since we are dealing with an octal string

codegeneration/csharp/Generator.js
aherlihy: 👍 java does this as well but also has an option to pass them individually. It's doing a bitwise-or, which I believe is what most other languages are doing under the surface as well.

symbols/csharp/templates.yaml
```diff
@@ -20,11 +20,8 @@ Templates:
     HexTypeTemplate: &HexTypeTemplate null
     OctalTypeTemplate: &OctalTypeTemplate !!js/function >
         (literal) => {
-          if ((literal.charAt(0) === '0' && literal.charAt(1) === '0') ||
-            (literal.charAt(0) === '0' && (literal.charAt(1) === 'o' || literal.charAt(1) === 'O'))) {
-            return `0${literal.substr(2, literal.length - 1)}`;
-          }
-          return literal;
+          literal = literal.replace(/[oO]+/g, '0')
+          return parseInt(literal, 8)
```
aherlihy: I think you probably want to conver this back into a string here explicitly, since all the template functions return strings.

test/json/success/javascript/js-constructors.json
```diff
@@ -173,39 +173,39 @@
...
+        "csharp": "new DateTime(1995, 12, 17, 3, 24, 0)",
         "shell": "Date('December 17, 1995 03:24:00Z')"
       },
       {
         "description": "Date string with UTC number",
         "javascript": "Date(819167040000)",
         "python": "datetime.datetime(1995, 12, 17, 2, 24, 0, tzinfo=datetime.timezone.utc).strftime('%a %b %d %Y %H:%M:%S %Z')",
         "java": "new java.util.Date(new java.lang.Long(819167040000)).toString()",
-        "csharp": "",
+        "csharp": "new DateTime(1995, 12, 17, 2, 24, 0)",
```
aherlihy: Does `new DateTime` return a string or a date object? This test is for just running `Date`, not `new Date`, so it should just return a string in the same format as the other languages.

symbols/csharp/templates.yaml
lrlna: oh didn't know that! will do

test/json/success/javascript/js-constructors.json
lrlna: yea`new DateTime()` always returns a `DateTime` object. The only way to get a string is to call a `.ToString` method, which is I guess what you did with java, so I'll do the same here ^__^

codegeneration/csharp/Generator.js
lrlna: let me take a look, maybe there is a way to do that! it's really annoying to do this appending thing, plus not all the things are available :/

lrlna: ok you can do the same thing, it's just a bit weird. i am trying to get both `?im` to work in an interpreter and it's beinggg weiiiiiiiiiirddd :O :O this may be a thing i ask craig about also

lrlna:
@aherlihy i've fixed the things you've pointed out!

lrlna:
going to be merging #54 instead, since it's all rebased with master :tada:

PR #51, opened 2018-05-01
Java minor fixes

aherlihy:
I ran every test java output, and found a few edge cases where it wouldn't compile.
- `org.bson.BsonBinarySubType.UUID --> org.bson.BsonBinarySubType.UUID_STANDARD`
- `Pattern.compile("ab\/cd") --> Pattern.compile("ab\\/cd")`
- All dates should be created from a `java.lang.Long(string)` so that we don't exceed the java integer size.
- All long's should be constructed from strings, except hex/decimal/octal so that we don't exceed that java integer size.
- The arguments to DBRef in java are `(coll, id)` or `(db, coll id)` instead of `(coll, id, db)` as in javascript.

NOTE: branched from "clean" so needs to be rebased merged after #49 . Currently being compared to that branch so the diff is easy to read.


codegeneration/java/Generator.js
```diff
@@ -268,6 +268,10 @@ module.exports = (superClass) => class ExtendedVisitor extends superClass {
     }
     return `new java.lang.Long(${doubleQuoteStringify(longstr)})`;
   }
+  //
```
lrlna: do you still want these few lines here? ^__^

symbols/java/templates.yaml
```diff
@@ -21,7 +21,11 @@ Templates:
         }
     LongBasicTypeTemplate: &LongBasicTypeTemplate !!js/function >
         (literal) => {
-            return `new java.lang.Long(${literal})`;
+            literal = literal.toString();
+            if (literal.indexOf('x') !== -1 || literal.indexOf('.') !== -1 || literal.charAt(0) === '0') {
+                return `new java.lang.Long(${literal})`;
```
lrlna: is it too early in the morning, or are you returning the same thing on line 28 without the `if` statement too?

codegeneration/java/Generator.js
aherlihy: No! I got rid of them in a later PR tho so it will be gone before it's merged

symbols/java/templates.yaml
aherlihy: Good catch!! Will update it

PR #55, opened 2018-05-02
Add support for compiling shell to javascript

aherlihy:
Some explanatory comments below, but it's pretty straightforward because 90% of the time JS and shell are the same.


codegeneration/javascript/Generator.js
```diff
@@ -3,4 +3,20 @@
  * For every input language, can "inherit" the generated visitNode methods.
  */
 module.exports = (superClass) => class ExtendedVisitor extends superClass {
+  constructor() {
+    super();
+  }
+  emitISODate(ctx) {
```
aherlihy: Needs to be special cased because in every other case, we want `new` to be the same as it is in shell, but here we always want `new Date(x)`.

codegeneration/javascript/Visitor.js
```diff
@@ -129,7 +131,7 @@ class Visitor extends ECMAScriptVisitor {
    * @return {String}
    */
   visitElision(ctx) {
-    ctx.type = this.Types._null;
+    ctx.type = this.Types._undefined;
```
aherlihy: This was a bug, elisions actually produce undefined and not null. Turns out javascript is the only language that even makes that distinction, so it's wasn't caught earlier.

test/json/success/shell/bson-constructors.json
```diff
@@ -347,34 +347,34 @@
...
         "shell": "[,,]"
       },
       {
         "description": "Array with elision in the middle",
-        "javascript": "[1,,,,2]",
+        "javascript": "[1, undefined, undefined, undefined, 2]",
         "python": "[1, None, None, None, 2]",
         "java": "Arrays.asList(new java.lang.Long(\"1\"), null, null, null, new java.lang.Long(\"2\"))",
-        "csharp": "new BsonArray {1, BsonNull.Value, BsonNull.Value, BsonNull.Value, 2}",
+        "csharp": "new BsonArray {1, BsonUndefined.Value, BsonUndefined.Value, BsonUndefined.Value, 2}",
```
aherlihy: Bugfix

codegeneration/javascript/Visitor.js
lrlna: oh whoa TIL

PR #57, opened 2018-05-02
Move JS specific code out of generators

aherlihy:
There was a bunch of TODO comments for moving anything JS specific out of the generators and into the visitor, and I finally got around to them! Except Binary and BinData, I didn't have time but it should be relatively straightforward to remove those as well.

The generators should be source language agnostic so that we can reuse them with visitors from any language. They should receive preprocessed info from the visitors so they don't care what the original code looked like.

I added an intermediate layer in between the `visitX` methods (Visitor), which visit tree nodes and always happen first, and the `emitX` methods (Generator) which usually call the templates and actually generate the code. The `processX` methods (Visitor) now are called from the `visit` methods. They will do anything input-language-specific, like executing code, but don't visit nodes.

The call order is now: `visitor.visitX --> visitor.processX --> generator.emitX --> templates` instead of `visitor.visitX --> generator.emitX --> templates`. I was hoping we could get rid of emit entirely and only use templates, but there are too many edge cases.

This is also nice because we had a bunch of repeated code in each generator that we can now get rid of.


codegeneration/csharp/Generator.js
```diff
@@ -271,13 +188,18 @@ module.exports = (superclass) => class ExtendedVisitor extends superclass {
...
+   * grandchild nodes: argumentList?
+   * great-grandchild nodes: singleExpression+
    *
-   * @returns {string} - DateTime(date)
+   * @param {FuncCallExpressionContext} ctx
+   * @param {Date} date
+   * @return {String}
    */
-  emitDate(ctx) {
+  emitDate(ctx, date) {
```
aherlihy: This is a bit of an edge case. The easiest way to provide date information is with a date object, so the "process" method executes the JS and creates a date object that gets passed to the "emit" methods.  The emit methods then extract whatever information they need to construct the generated code. With other input languages we can probably also create date objects.

codegeneration/javascript/Visitor.js
```diff
@@ -438,6 +436,248 @@ class Visitor extends ECMAScriptVisitor {
...
+      return this.emitRegExp(ctx, pattern, targetflags);
+    }
+
+    if (ctx.type.template) {
+      return ctx.type.template(pattern, targetflags);
+    }
+    return this.visitChildren(ctx);
+  }
+
+  process_regex(ctx) {
```
lrlna: is there something special about this method that it's snake-cased?

codegeneration/javascript/Visitor.js
```diff
@@ -438,6 +436,248 @@ class Visitor extends ECMAScriptVisitor {
...
+      longstr = this.executeJavascript(`new ${ctx.getText()}`).toString();
+    } catch (error) {
+      throw new SemanticGenericError({message: error.message});
+    }
+    if ('emitLong' in this) {
+      return this.emitLong(ctx, longstr);
+    }
+    const lhs = symbolType.template ? symbolType.template() : 'Long';
+    const rhs = symbolType.argsTemplate ? symbolType.argsTemplate(lhs, longstr) : `(${longstr})`;
+    return `${this.new}${lhs}${rhs}`;
```
lrlna: i can't figure out how the visitor is going to know whether a language needs a `new` keyword or not, since it's still set in a generator which is a "level-down" in the inheritance tree.. am i missing something in the code here?

codegeneration/javascript/Visitor.js
aherlihy: no, that was just old habits, will change it now!

codegeneration/javascript/Visitor.js
aherlihy: I'm not sure I understand your question - super classes can access subclasses fields, like how the visitor is calling `emit` methods that are defined on the subclass.

Languages like java, that always require a `new` for constructors, will set `this.new` to a string so that it gets added when it's needed. We don't need to check if the symbol is a constructor type because we already know what the symbol is, in this example `Long` which is a constructor type. 

If the language always needs `new` for constructors, but for some reason not this one, then the generator can add an `emitX` method that either drops or adds a `new`. For example, java/generator.js:emitDecimal128.

codegeneration/javascript/Visitor.js
aherlihy: Oh sorry no, actually, that is required. Because regex is a base type, the name of the type is `_regex`. So `process` + `_regex` = `process_regex`.

lrlna: is just the `regex` already taken as a type? or does the `_` signify something in `_regex` ? Cause all the other types are just `Date`, `Long` etc

codegeneration/javascript/Visitor.js
lrlna: cool gotcha!

codegeneration/javascript/Visitor.js
lrlna: ohhhh gotcha! ok that makes sense, ty ty

PR #59, opened 2018-05-03
c# bson-utils

lrlna:
this pr adds bson-utils tests and templates for c#. The larger diff is due to space changes in the yaml file to follow the 4-space convention for yaml that i wasn't following before.

a really large thing to still double check is how to handle the `Long` to `BsonInt64` conversions. I am still not exactly sure when to use `BsonInt64.ToInt64()` vs `Convert.ToInt64()` vs. `Int64.Parse()`. That's something I'll double check with Craig, but that will be done after this merge.


durran:
YAML indentation can be whatever one likes, but the convention is 2 spaces not 4.

lrlna:
@durran ah sorry, to elaborate, i meant the convention we have with all the other `.yml` files. but it's also good to know it's usually 2 spaces!

lrlna:
:package: `0.1.0`

PR #60, opened 2018-05-04
c# bson object methods

lrlna:
Adds currently applicable bson object methods. There are a few things that will end up in a merge conflict with the utils branch, as I've decided to keep all `Long` conversions to `new BsonInt64.ToInt64(val)`. It elimiates the whole parsing from int 32 woo o/ 

Something I am still not clear about with c# is whether or not Properties (these are thigns like `.Value`, `AsBsonSymbol`) are callable in the language, so i am keeping them as is for now until i get confirmation.


lrlna:
:package: `0.1.0`

PR #61, opened 2018-05-04
[WIP] verify c# output

lrlna:
This branch is meant to create a good-ish readable diff to our tests for c# so the output can be verified. If changes are to be made, the result will be merged and rebased into master instead of `remove-c-sharp-tests` branch.

The main chunk of verification should happen in the `test/json/success/javascript`, but everything else is fair game too.

I will comment out the code I have specific questions about, which will hopefully make this easier to read. Some will be language specific (never written c# in my life before this), and the others are driver specific, such as things I couldn't quite figure out from documentation.

The TL;DR of what these tests are doing is checking whether `javascript` code is correctly generated for the applicable language (this case `c#`).


lrlna:
ugh i fucked up a rebase; gonna close and open a new one.

PR #65, opened 2018-05-17
[WIP] verify c# tests

lrlna:
(third time is a charm)

This branch is meant to create a good-ish readable diff to our tests for `c#` so the output can be verified. The main chunk of verification should happen in the t`est/json/success/javascript`, but everything else is fair game too.

I will comment out the code I have specific questions about, which will hopefully make this easier to read. Some will be language specific (never written `c#` in my life before this), and the others are driver specific, such as things I couldn't quite figure out from documentation.

The TL;DR of what these tests are doing is checking whether javascript code is correctly generated for the applicable language (this case `c#`).

I couldn't get the PR to work properly the proper way around, so the **RED** represents the c# output (rather than the expected green).

### DIFF
--- a/test/json/success/javascript/bson-constructors.json
+++ b/test/json/success/javascript/bson-constructors.json
@@ -6,7 +6,7 @@
         "javascript": "new Code('some code')",
         "python": "Code('some code')",
         "java": "new Code(\"some code\")",
-        "csharp": "new BsonJavaScript(@\"some code\")",
+        "csharp": "",
         "shell": "Code('some code')"
       },
       {
@@ -14,7 +14,7 @@
         "javascript": "Code('some code')",
         "python": "Code('some code')",
         "java": "new Code(\"some code\")",
-        "csharp": "new BsonJavaScript(@\"some code\")",
+        "csharp": "",
         "shell": "Code('some code')"
       },
       {
@@ -22,7 +22,7 @@
         "javascript": "Code('string', {x: '1'})",
         "python": "Code('string', {'x': '1'})",
         "java": "new CodeWithScope(\"string\", new Document().append(\"x\", \"1\"))",
-        "csharp": "new BsonJavaScriptWithScope(@\"string\", new BsonDocument(\"x\", \"1\"))",
+        "csharp": "",
         "shell": "Code('string', {x: '1'})"
       },
       {
@@ -30,7 +30,7 @@
         "javascript": "Code(function(test) { console.log(test); })",
         "python": "Code('function(test){console.log(test);}')",
         "java": "new Code(\"function(test){console.log(test);}\")",
-        "csharp": "new BsonJavaScript(@\"function(test){console.log(test);}\")",
+        "csharp": "",
         "shell": "Code('function(test){console.log(test);}')"
       },
       {
@@ -38,7 +38,7 @@
         "javascript": "new Code('this.a > i', { i: '1' })",
         "python": "Code('this.a > i', {'i': '1'})",
         "java": "new CodeWithScope(\"this.a > i\", new Document().append(\"i\", \"1\"))",
-        "csharp": "new BsonJavaScriptWithScope(@\"this.a > i\", new BsonDocument(\"i\", \"1\"))",
+        "csharp": "",
         "shell": "Code('this.a > i', {i: '1'})"
       }
     ],
@@ -48,7 +48,7 @@
         "javascript": "ObjectId()",
         "python": "ObjectId()",
         "java": "new ObjectId()",
-        "csharp": "new ObjectId()",
+        "csharp": "",
         "shell": "ObjectId()"
       },
       {
@@ -56,7 +56,7 @@
         "javascript": "new ObjectId()",
         "python": "ObjectId()",
         "java": "new ObjectId()",
-        "csharp": "new ObjectId()",
+        "csharp": "",
         "shell": "new ObjectId()"
       },
       {
@@ -64,7 +64,7 @@
         "javascript": "ObjectId('5a7382114ec1f67ae445f778')",
         "python": "ObjectId('5a7382114ec1f67ae445f778')",
         "java": "new ObjectId(\"5a7382114ec1f67ae445f778\")",
-        "csharp": "new ObjectId(\"5a7382114ec1f67ae445f778\")",
+        "csharp": "",
         "shell": "new ObjectId('5a7382114ec1f67ae445f778')"
       }
     ],
@@ -74,7 +74,7 @@
         "javascript": "Binary(Buffer.from('1234'))",
         "python": "Binary(b'1234')",
         "java": "new Binary(\"1234\".getBytes(\"UTF-8\"))",
-        "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"))",
+        "csharp": "",
         "shell": "new BinData(4, '1234')"
       },
       {
@@ -82,7 +82,7 @@
         "javascript": "new Binary(Buffer.from(\"1234\"), Binary.SUBTYPE_UUID)",
         "python": "Binary(b'1234', bson.binary.UUID_SUBTYPE)",
         "java": "new Binary(org.bson.BsonBinarySubType.UUID_STANDARD, \"1234\".getBytes(\"UTF-8\"))",
-        "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"), BsonBinarySubType.UuidStandard)",
+        "csharp": "",
         "shell": "new BinData(4, '1234')"
       },
       {
@@ -90,7 +90,7 @@
         "javascript": "Binary(Buffer.from('1234'), '1')",
         "python": "Binary(b'1234', bson.binary.FUNCTION_SUBTYPE)",
         "java": "new Binary(org.bson.BsonBinarySubType.FUNCTION, \"1234\".getBytes(\"UTF-8\"))",
-        "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"), BsonBinarySubType.Function)",
+        "csharp": "",
         "shell": "new BinData(1, '1234')"
       },
       {
@@ -98,7 +98,7 @@
         "javascript": "Binary('1234')",
         "python": "Binary(b'1234')",
         "java": "new Binary(\"1234\".getBytes(\"UTF-8\"))",
-        "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"))",
+        "csharp": "",
         "shell": "new BinData(1, '1234')"
       },
       {
@@ -106,7 +106,7 @@
         "javascript": "Binary('1234', '1')",
         "python": "Binary(b'1234', bson.binary.FUNCTION_SUBTYPE)",
         "java": "new Binary(org.bson.BsonBinarySubType.FUNCTION, \"1234\".getBytes(\"UTF-8\"))",
-        "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"), BsonBinarySubType.Function)",
+        "csharp": "",
         "shell": "new BinData(4, '1234')"
       }
     ],
@@ -116,7 +116,7 @@
         "javascript": "new DBRef('coll', new ObjectId())",
         "python": "DBRef('coll', ObjectId())",
         "java": "new DBRef(\"coll\", new ObjectId())",
-        "csharp": "new MongoDBRef(\"coll\", new ObjectId())",
+        "csharp": "",
         "shell": "new DBRef('coll', new ObjectId())"
       },
       {
@@ -124,7 +124,7 @@
         "javascript": "new DBRef('coll', ObjectId(), 'db')",
         "python": "DBRef('coll', ObjectId(), 'db')",
         "java": "new DBRef(\"db\", \"coll\", new ObjectId())",
-        "csharp": "new MongoDBRef(\"coll\", new ObjectId(), \"db\")",
+        "csharp": "",
         "shell": "new DBRef('coll', ObjectId(), 'db')"
       }
     ],
@@ -134,7 +134,7 @@
         "javascript": "Int32(3)",
         "python": "int(3)",
         "java": "new java.lang.Integer(3)",
-        "csharp": "Convert.ToInt32(3)",
+        "csharp": "",
         "shell": "NumberInt(3)"
       },
       {
@@ -142,7 +142,7 @@
         "javascript": "Int32('3')",
         "python": "int('3')",
         "java": "new java.lang.Integer(\"3\")",
-        "csharp": "Int32.Parse(\"3\")",
+        "csharp": "",
         "shell": "NumberInt('3')"
       }
     ],
@@ -152,7 +152,7 @@
         "javascript": "Double(3)",
         "python": "float(3)",
         "java": "new java.lang.Double(3)",
-        "csharp": "new BsonDouble(3)",
+        "csharp": "",
         "shell": "3"
       },
       {
@@ -160,7 +160,7 @@
         "javascript": "new Double(3)",
         "python": "float(3)",
         "java": "new java.lang.Double(3)",
-        "csharp": "new BsonDouble(3)",
+        "csharp": "",
         "shell": "3"
       },
       {
@@ -168,7 +168,7 @@
         "javascript": "Double('3')",
         "python": "float('3')",
         "java": "new java.lang.Double(\"3\")",
-        "csharp": "new BsonDouble(3)",
+        "csharp": "",
         "shell": "3"
       }
     ],
@@ -178,7 +178,7 @@
         "javascript": "new Long(-1, 2147483647)",
         "python": "Int64(9223372036854775807)",
         "java": "new java.lang.Long(\"9223372036854775807\")",
-        "csharp": "new BsonInt64(9223372036854775807)",
+        "csharp": "",
         "shell": "new Long(9223372036854775807)"
       }
     ],
@@ -188,7 +188,7 @@
         "javascript": "new Decimal128(Buffer.from('5'))",
         "python": "Decimal128('5.3E-6175')",
         "java": "Decimal128.parse(\"5.3E-6175\")",
-        "csharp": "new Decimal128(5)",
+        "csharp": "",
         "shell": "new NumberDecimal(5.3E-6175)"
       }
     ],
@@ -198,7 +198,7 @@
         "javascript": "MinKey()",
         "python": "MinKey()",
         "java": "new MinKey()",
-        "csharp": "BsonMinKey.Value",
+        "csharp": "",
         "shell": "MinKey()"
       },
       {
@@ -206,7 +206,7 @@
         "javascript": "MaxKey()",
         "python": "MaxKey()",
         "java": "new MaxKey()",
-        "csharp": "BsonMaxKey.Value",
+        "csharp": "",
         "shell": "MaxKey()"
       },
       {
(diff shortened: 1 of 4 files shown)


test/json/success/javascript/bson-constructors.json
```diff
@@ -6,39 +6,39 @@
         "javascript": "new Code('some code')",
         "python": "Code('some code')",
         "java": "new Code(\"some code\")",
-        "csharp": "new BsonJavaScript(@\"some code\")",
```
lrlna: want to double check on this `@` sign before a string: is this common practice for all strings I should adhere to?

test/json/success/javascript/bson-constructors.json
```diff
@@ -178,7 +178,7 @@
         "javascript": "new Long(-1, 2147483647)",
         "python": "Int64(9223372036854775807)",
         "java": "new java.lang.Long(\"9223372036854775807\")",
-        "csharp": "new BsonInt64.ToInt64(9223372036854775807)",
```
lrlna: Sort of a similar question to the one about `Double`, should I provide `BsonInt64` a specific type, or would just the just-in-case conversion of `ToInt64` be okay here?

test/json/success/javascript/bson-constructors.json
```diff
@@ -134,15 +134,15 @@
         "javascript": "Int32(3)",
         "python": "int(3)",
         "java": "new java.lang.Integer(3)",
-        "csharp": "Convert.ToInt32(3)",
```
lrlna: Numbers have been especially confusing for me. I want to verify how people `Int32`/`Double` in real life. My thought process with `Int32` was that if I am dealing with an initial input that's an integer, I'd use `Convert.ToInt32(integer)`, but if it's a string, I'd use the `Int32.Parse()`. It'd be good to know if there is a single method people normally use.

test/json/success/javascript/bson-constructors.json
```diff
@@ -134,15 +134,15 @@
...
+        "csharp": "",
         "shell": "NumberInt(3)"
       },
       {
         "description": "Int32 with valid string arg",
         "javascript": "Int32('3')",
         "python": "int('3')",
         "java": "new java.lang.Integer(\"3\")",
-        "csharp": "Int32.Parse(\"3\")",
```
lrlna: kind of related to the above -- should i just use `Convert.ToInt32(3)` here?

test/json/success/javascript/bson-constructors.json
```diff
@@ -152,23 +152,23 @@
         "javascript": "Double(3)",
         "python": "float(3)",
         "java": "new java.lang.Double(3)",
-        "csharp": "new BsonDouble(\"3\")",
```
lrlna: To double check, do I always have to make sure that `BsonDouble` takes a specific `int`? or just passing in a string is something most people would do?

test/json/success/javascript/bson-constructors.json
```diff
@@ -268,55 +268,55 @@
...
-        "csharp": "new BsonDocument()",
+        "csharp": "",
         "shell": "{}"
       },
       {
         "description": "Two items in document",
         "javascript": "{x: '1', n: '4'}",
         "python": "{'x': '1', 'n': '4'}",
         "java": "new Document().append(\"x\", \"1\").append(\"n\", \"4\")",
-        "csharp": "new BsonDocument{ { \"x\", \"1\" }, { \"n\", \"4\" } }",
```
lrlna: with `BsonDocument` I've been using the strategy of if there are multiple `key/val` pairs, I use `{}`; and if it's just a single `key/val`, it's a `()`. Is that what most people do? or does everyone always use `{}`?

test/json/success/javascript/bson-methods.json
```diff
@@ -154,207 +154,207 @@
         "javascript": "Long(1, 100).toInt()",
         "python": "int(Int64(429496729601))",
         "java": "new java.lang.Long(\"429496729601\").intValue()",
-        "csharp": "new BsonInt64.ToInt64(429496729601).ToInt32()",
```
lrlna: this one is a bit weird and i wasn't sure whether I should stick with the above or something like this: `Convert.ToInt32(new BsonInt64.ToInt64(429496729601))`

test/json/success/javascript/bson-methods.json
```diff
@@ -154,207 +154,207 @@
...
+        "csharp": "",
         "shell": "NumberLong(429496729601).valueOf()"
       },
       {
         "description": "toNumber",
         "javascript": "Long(1, 100).toNumber()",
         "python": "float(Int64(429496729601))",
         "java": "new java.lang.Long(\"429496729601\").floatValue()",
-        "csharp": "new BsonInt64.ToInt64(429496729601).Value",
```
lrlna: is there a specific way to get a float value with an `int64`? can't for the life of me find this in the docs

test/json/success/javascript/bson-utils.json
```diff
@@ -99,71 +99,71 @@
...
-        "csharp": "new BsonInt64.ToInt64(\"5\")"
+        "csharp": ""
       },
       {
         "description": "fromString with radix",
         "javascript": "Long.fromString('5', 10)",
         "python": "Int64(int('5'))",
         "java": "java.lang.Long.parseLong(\"5\", 10)",
         "shell": "NumberLong('5')",
-        "csharp": "Int64.TryParse(\"5\", 10)"
```
lrlna: I wasn't able to figure out if `BsonInt64()` can take a radix as the parameter. From the docs it seems it doesn't, so I am using the `Int64.TryParse()` to be able to pass that in. Did I miss something? Is this normally how developers would work with `int64` and radix situation?

test/json/success/javascript/bson-constructors.json
```diff
@@ -116,15 +116,15 @@
         "javascript": "new DBRef('coll', new ObjectId())",
         "python": "DBRef('coll', ObjectId())",
         "java": "new DBRef(\"coll\", new ObjectId())",
-        "csharp": "new MongoDBRef(\"coll\", new ObjectId())",
```
lrlna: This seems to be the types the docs are asking for, but I am ending up with an error I am not 100% certain about:

```shell
Program.cs(30,20): error CS1503: Argument 1: cannot convert from 'string' to 'System.Collections.Generic.Dictionary<string, object>' [/Users/lrlna/code/csharp-playground/csharp-playground.csproj]
Program.cs(30,62): error CS1503: Argument 2: cannot convert from 'MongoDB.Driver.MongoDBRef' to 'System.Collections.Generic.IEnumerable<string>' [/Users/lrlna/code/csharp-playground/csharp-playground.csproj]
```
Is there anything I should be doing differently here, would you know?

test/json/success/javascript/bson-methods.json
```diff
@@ -154,207 +154,205 @@
...
-        "csharp": "new BsonInt64(429496729601) < 0",
+        "csharp": "",
         "shell": "NumberLong(429496729601) < 0"
       },
       {
         "description": "isOdd",
         "javascript": "Long(1, 100).isOdd()",
         "python": "(Int64(429496729601) % 2) == 0",
         "java": "new java.lang.Long(\"429496729601\") % 2 == 0",
-        "csharp": "new BsonInt64(429496729601).ToInt64() % 2 == 0",
```
lrlna: Sorta of a related question to the one above about numbers. I wasn't able to compile this as `new BsonInt64(429496729601) % 2 == 0` so I had to add a `.ToInt64` method to get this to work, and I am not sure whether this even makes sense?

lrlna: these are the kinds of errors i got:

```shell
Operator '%' cannot be applied to operands of type 'BsonInt64' and 'int'
Operator '+' cannot be applied to operands of type 'BsonInt64' and 'BsonInt64'
```

test/json/success/javascript/bson-utils.json
```diff
@@ -7,22 +7,22 @@
...
-        "csharp": "new ObjectId(\"5ab901c29ee65f5c8550c5b9\")"
+        "csharp": "
       },
       {
         "description": "createFromTime",
         "javascript": "ObjectId.createFromTime(1522075031642)",
         "python": "ObjectId.from_datetime(datetime.datetime.fromtimestamp(1522075031642/1000.0))",
         "java": "new ObjectId(new java.util.Date(new java.lang.Long(\"1522075031642\")))",
-        "shell": "ObjectId.fromDate(new Date(1522075031642))",
-        "csharp": "new ObjectId().GenerateNewId(Convert.ToInt32(1522075031642))"
```
lrlna: had to hack this one a bit so it compiles by explicitly converting to an `int32` which seems to be fine. 

however, i don't think i am accessing this method properly, as the compiler gives me this back:

```shell
Member 'ObjectId.GenerateNewId(int)' cannot be accessed with an instance reference; qualify it with a type name instead
```
i guess i'd have to define a new instance of `ObjectId()` separately and then get this to work?

test/json/success/javascript/bson-constructors.json
craiggwilson: Using `@` is only necessary when you want to escape sequences to be ignored, which is not normally the case. I'd suggest removing it.

test/json/success/javascript/bson-constructors.json
craiggwilson: .NET has implicit conversions. So even though BsonInt32 is what will happen underneath, we don't really need to use those. And since .NET has literals, then just using `3` would be sufficient.

test/json/success/javascript/bson-constructors.json
craiggwilson: Same here... just use `3`

test/json/success/javascript/bson-constructors.json
craiggwilson: Most people probably just use `{` everywhere.

test/json/success/javascript/bson-constructors.json
```diff
@@ -152,23 +152,23 @@
         "javascript": "Double(3)",
         "python": "float(3)",
         "java": "new java.lang.Double(3)",
-        "csharp": "new BsonDouble(3)",
```
craiggwilson: `3.0`

test/json/success/javascript/bson-constructors.json
craiggwilson: Hmm... I'm not sure what the issue is here. Seems like it should work just fine. Can I see the program that was generated that is actually causing this issue?

test/json/success/javascript/bson-constructors.json
```diff
@@ -152,23 +152,23 @@
...
+        "csharp": "",
         "shell": "3"
       },
       {
         "description": "new Double with number arg",
         "javascript": "new Double(3)",
         "python": "float(3)",
         "java": "new java.lang.Double(3)",
-        "csharp": "new BsonDouble(3)",
```
craiggwilson: `3.0`

test/json/success/javascript/bson-constructors.json
```diff
@@ -152,23 +152,23 @@
...
+        "csharp": "",
         "shell": "3"
       },
       {
         "description": "Double with valid string arg",
         "javascript": "Double('3')",
         "python": "float('3')",
         "java": "new java.lang.Double(\"3\")",
-        "csharp": "new BsonDouble(3)",
```
craiggwilson: `3.0`

test/json/success/javascript/bson-constructors.json
```diff
@@ -178,7 +178,7 @@
         "javascript": "new Long(-1, 2147483647)",
         "python": "Int64(9223372036854775807)",
         "java": "new java.lang.Long(\"9223372036854775807\")",
-        "csharp": "new BsonInt64(9223372036854775807)",
```
craiggwilson: `9223372036854775807L`

test/json/success/javascript/bson-constructors.json
```diff
@@ -402,15 +402,15 @@
         "javascript": "Symbol('2')",
         "python": "unicode('2', 'utf-8')",
         "java": "new Symbol(\"2\")",
-        "csharp": "new BsonString(\"2\")",
```
craiggwilson: `"2"`

test/json/success/javascript/bson-constructors.json
```diff
@@ -402,15 +402,15 @@
...
+        "csharp": "",
         "shell": "'2'"
       },
       {
         "description": "new Symbol from string",
         "javascript": "new Symbol('2')",
         "python": "unicode('2', 'utf-8')",
         "java": "new Symbol(\"2\")",
-        "csharp": "new BsonString(\"2\")",
```
craiggwilson: `"2"`

test/json/success/javascript/bson-methods.json
```diff
@@ -67,31 +67,31 @@
         "javascript": "Binary(Buffer.from('a string')).length()",
         "python": "len(bytearray(Binary(b'a string')))",
         "java": "new Binary(\"a string\".getBytes(\"UTF-8\")).length()",
-        "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"a string\")).Length",
```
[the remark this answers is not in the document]
craiggwilson: `new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"a string\")).Bytes.Length`

test/json/success/javascript/bson-methods.json
```diff
@@ -135,7 +135,7 @@
         "javascript": "Int32(3).valueOf()",
         "python": "int(3)",
         "java": "new java.lang.Integer(3).intValue()",
-        "csharp": "Convert.ToInt32(3)"
```
craiggwilson: `3`

test/json/success/javascript/bson-methods.json
```diff
@@ -144,7 +144,7 @@
         "javascript": "Double(3).valueOf()",
         "python": "float(3)",
         "java": "new java.lang.Double(3).doubleValue()",
-        "csharp": "new BsonDouble(3).Value",
```
craiggwilson: `3.0`

test/json/success/javascript/bson-methods.json
```diff
@@ -154,207 +154,205 @@
         "javascript": "Long(1, 100).toInt()",
         "python": "int(Int64(429496729601))",
         "java": "new java.lang.Long(\"429496729601\").intValue()",
-        "csharp": "new BsonInt64(429496729601).ToInt32()",
```
craiggwilson: `(int)429496729601`

craiggwilson: I'm not totally sure what we are testing here...

test/json/success/javascript/bson-methods.json
```diff
@@ -154,207 +154,205 @@
...
-        "csharp": "\"429496729601\"",
+        "csharp": "",
         "shell": ""
       },
       {
         "description": "isZero",
         "javascript": "Long(1, 100).isZero()",
         "python": "Int64(429496729601) == 0",
         "java": "new java.lang.Long(\"429496729601\").equals(new java.lang.Long(0))",
-        "csharp": "new BsonInt64(429496729601) == 0",
```
craiggwilson: 429496729601L == 0

test/json/success/javascript/bson-methods.json
```diff
@@ -154,207 +154,205 @@
...
+        "csharp": "",
         "shell": "NumberLong(429496729601) === 0"
       },
       {
         "description": "isNegative",
         "javascript": "Long(1, 100).isNegative()",
         "python": "Int64(429496729601) < 0",
         "java": "new java.lang.Long(\"429496729601\").compareTo(new java.lang.Long(0)) < 0",
-        "csharp": "new BsonInt64(429496729601) < 0",
```
craiggwilson: 429496729601L < 0

test/json/success/javascript/bson-methods.json
craiggwilson: these types of operators aren't defined on the BsonXXX types. You don't really need them, although I'm unsure of how this code is getting used. Typically, you'd just do this: `429496729601L % 2`. Having a full example of how this is getting used would help.

test/json/success/javascript/bson-utils.json
craiggwilson: `ObjectID.GenerateNewId(1522075031642)` should work.

test/json/success/javascript/language-types.json
```diff
@@ -164,87 +164,87 @@
...
-        "csharp": "new DateTime(1995, 12, 17, 3, 24, 0).ToString()",
+        "csharp": "",
         "shell": "Date(1995, 11, 17, 3, 24, 0)"
       },
       {
         "description": "Current time",
         "javascript": "Date.now()",
         "python": "datetime.datetime.utcnow()",
         "java": "new java.util.Date()",
-        "csharp": "DateTime.Now",
```
craiggwilson: `DateTime.UtcNow`... This will work, but should probably emphasize that this will get stored as UTC and not roundtrip as a local date.

test/json/success/javascript/bson-constructors.json
lrlna: amazing! this is good news ✨

test/json/success/javascript/bson-constructors.json
lrlna: It's not the greatest of programs, but I generate the file based on the output from my tests, and it ends up looking like this:

```csharp
using MongoDB.Bson;
using MongoDB.Driver;

using System;
using System.Text.RegularExpressions;

namespace csharp_test
{
    class Program
    {
        static void Main(string[] args)
        {
            var bsonconstructorsDBRef = new BsonDocument{
                  {"new DBRef with (string, ObjectID) args", new MongoDBRef("coll", new ObjectId())},
                  {"new DBRef with (string, ObjectId, string) args", new MongoDBRef("coll", new ObjectId(), "db")}
            };
            Console.WriteLine("Hello World!");
         }
    }
}

test/json/success/javascript/bson-methods.json
lrlna: > these types of operators aren't defined on the BsonXXX types
ohh ok cool. 

I don't have a particular example for how this code will be used to be honest, this particular conversion came straight up from trying to find an appropriate matching method to javascript's. 

I think I will use the suggestion of `429496729601L % 2` since that seems a lot more readable.

lrlna: and a follow up question, if I just use `429496729601L` instead of the the `BsonInt64`, I should be able to do `!429496729601L` and `429496729601L << 10` for negation and bit shifting, right?

test/json/success/javascript/bson-constructors.json
craiggwilson: Ok, so MongoDBRef cannot be used in a BsonDocument like this. There isn't any implicit conversions defined for this type. You'll just need to compose the reference manually as a BsonDocument.

test/json/success/javascript/bson-methods.json
craiggwilson: You can use it for bit shifting, yes. Negation uses the hat operator `^`.

test/json/success/javascript/bson-utils.json
craiggwilson: `new` creates  a new instance. In this case, you are calling a static method on the ObjectId type, which doesn't require an instance.

test/json/success/javascript/bson-constructors.json
lrlna: right that's fair! thank you.

test/json/success/javascript/bson-methods.json
lrlna: cool cool thank you!

lrlna:
@craiggwilson two more questions i think!

1. Is there a way to get a `Decimal128()` from a string? or is it just an `int` situation?

2. In javascript you can create a `Long` from string and provide a radix. Is there a good way of doing that in c#?

lrlna:
closing this one up, since I think everything's been resolved. Thank you so much for your help, Craig!

PR #66, opened 2018-05-17
Some validation cases ideas

alenakhineika:
The motivation of current PR is below.


lrlna:
@alenakhineika could you write out the changes you're proposing to make in this PR so I can read through the code better? I am seeing a few new `emit` methods and a few new error methods and it would be good to have a summary of what's what ^___^

lrlna:
@alenakhineika could you write out the changes you're proposing to make in this PR so I can read through the code better? I am seeing a few new `emit` methods and a few new error methods and it would be good to have a quick summary of what's what ^___^

alenakhineika:
The motivation of current PR is creating tests for error cases and providing consistency for validation and throwing error types.

Currently some of our constructors throw an SemanticArgumentCountMismatchError and SemanticTypeError errors, eg Int32(). But some of them, which have emit methods throw only SemanticGenericError error. My idea, that all constructors should behave the same way. Therefore I have updated eg processBinaryFromJS function with some validation logic and throw errors according to it.

I have changed only couple of constructors to let you check the idea. If everyone is ok with this approach i will proceed with rest of constructors.

codegeneration/javascript/Visitor.js
```diff
@@ -553,18 +629,30 @@ class Visitor extends ECMAScriptVisitor {
    */
   processLong(ctx) {
     ctx.type = this.Types.Long;
+
     const symbolType = this.Symbols.Long;
+    const argList = ctx.arguments().argumentList();
```
aherlihy: We can use the `checkArguments` function here.

codegeneration/javascript/Visitor.js
```diff
@@ -525,13 +585,29 @@ class Visitor extends ECMAScriptVisitor {
...
     const symbolType = this.Symbols.ObjectId;
     const argList = ctx.arguments().argumentList();
     const lhs = symbolType.template ? symbolType.template() : 'ObjectId';
+
     if (!argList) {
       return `${this.new}${lhs}()`;
     }
-    let hexstr;
+
+    const args = argList.singleExpression();
```
aherlihy: We can use the `checkArguments` method here instead of checking by hand. We also need to check that it's a valid hex string, unless we want to leave that error to the execution.

symbols/basic_types.yaml
```diff
@@ -55,4 +55,7 @@ BasicTypes:
         <<: *__type
         id: "_undefined"
         template: *UndefinedTypeTemplate
-
+    _buffer: &BufferType
```
aherlihy: Do we want to actually support the buffer type outside of creating a Binary object? I think we probably want to stay away from adding new types, since then we have to add support for translation between all languages. I'm okay with only allowing `Buffer` to be viable within a Binary constructor in javascript, since I expect 99% of the time people are going to be constructing Binary objects with the utils and not the constructor.

test/helpers.js
```diff
@@ -52,12 +60,11 @@ const checkResults = {
 
   error: function(inputLang, outputLang, test) {
     try {
-      compiler[outputLang](test[inputLang]);
+      compiler[inputLang][outputLang](test.query);
+      expect(1).to.equal(0);
```
aherlihy: I don't think this is needed if we add an expect-to-throw clause. If we do need to fail, we can add an assert(false) outside of the try-catch, but we don't want to have it here.

codegeneration/javascript/Visitor.js
```diff
@@ -623,7 +711,7 @@ class Visitor extends ECMAScriptVisitor {
     if ('emitNumberDecimal' in this) {
       return this.emitNumberDecimal(ctx, decstr);
     }
-    const lhs = symbolType.template ? symbolType.template() : 'NumberDecimal';
+    const lhs = symbolType.template ? symbolType.template() : 'Decimal128';
```
aherlihy: This is within `processNumberDecimal`, so the type has to be `NumberDecimal`. Decimal128 does not exist of we are accepting NumberDecimal. We can move the processX methods for the shell types (i.e. processNumberDecimal, processNumberLong, etc.) to the shell/visitor **in a separate PR**.

codegeneration/javascript/Visitor.js
```diff
@@ -471,6 +473,64 @@ class Visitor extends ECMAScriptVisitor {
     return this.processRegExp(ctx);
   }
 
+  processBufferFrom(ctx) {
```
aherlihy: this should just be processBuffer

codegeneration/javascript/Visitor.js
```diff
@@ -471,6 +473,64 @@ class Visitor extends ECMAScriptVisitor {
...
+    return ctx;
+  }
+
+  processBinaryFromJS(ctx) {
+    ctx.type = this.Types.Binary;
+
+    const argList = ctx.arguments().argumentList();
+    let bytes = '';
+
+    if (
```
aherlihy: You can use the existing checkArguments method here instead of doing it by hand.

codegeneration/javascript/Visitor.js
```diff
@@ -471,6 +473,64 @@ class Visitor extends ECMAScriptVisitor {
...
+      throw new SemanticArgumentCountMismatchError({
+        message: 'Binary requires one or two arguments'
+      });
+    }
+
+    const args = argList.singleExpression();
+
+    bytes = this.visit(args[0]);
+
+    if (
```
aherlihy: This is also covered by checkArguments.

codegeneration/python/Generator.js
```diff
@@ -179,27 +183,21 @@ module.exports = (superClass) => class ExtendedVisitor extends superClass {
...
-    try {
-      binobj = this.executeJavascript(ctx.getText());
-      type = binobj.sub_type;
-    } catch (error) {
-      throw new SemanticGenericError({message: error.message});
-    }
-
-    const bytes = singleQuoteStringify(binobj.toString());
-    const argList = ctx.arguments().argumentList().singleExpression();
+  emitBinaryFromJS(ctx, bytes, subtype) {
```
aherlihy: If the processBinary method is written correctly, then all the emitBinary methods should have the same arguments regardless of if it came from shell or javascript.

symbols/csharp/templates.yaml
```diff
@@ -595,6 +596,7 @@ Templates:
     SymbolSymbolArgsTemplate: &SymbolSymbolArgsTemplate null
     BSONRegExpSymbolTemplate: &BSONRegExpSymbolTemplate null # Has emit method
     BSONRegExpSymbolArgsTemplate: &BSONRegExpSymbolArgsTemplate null # Has emit method
+    Decimal128TypeTemplate: &Decimal128TypeTemplate null
```
aherlihy: I don't think we need Decimal128TypeTemplate. There are only `TypeTemplate`s (as opposed to `SymbolTemplate`) for the basic types. These need type templates because they're going from a literal, while Decimal128 can never be a literal. Decimal is a literal, because you can write `1.0001`, but you cannot ever create a Decimal128 without a function call, which means that the template is covered by the Symbol template.

The other languages, c# and java, are able to handle Decimal128 without needing a type template, which is another indicator that we don't need to add it.

test/helpers.js
```diff
@@ -2,19 +2,27 @@ const parse = require('fast-json-parse');
...
 // Need a way to have test pass while developing
+const inputLanguages = ['javascript', 'shell'];
 const outputLanguages = ['csharp', 'python', 'java', 'javascript'];
-const inputLanguages = ['shell', 'javascript'];
+
 const unsupported = {
   success: {
     javascript: {
-      java: {},
+      java: {
```
aherlihy: I'm not sure why this would need to break the tests for all the other languages.

test/json/error/javascript/bson-methods.json
```diff
@@ -0,0 +1,89 @@
+{
```
aherlihy: You should add some tests for actually testing the bson methods themselves, not just the attribute access.

codegeneration/javascript/Visitor.js
```diff
@@ -719,19 +722,32 @@ class Visitor extends ECMAScriptVisitor {
...
+      throw new SemanticArgumentCountMismatchError({
+        message: 'Binary requires one or two arguments'
+      });
+    }
+
     try {
       binobj = this.executeJavascript(ctx.getText());
       type = binobj.sub_type;
     } catch (error) {
       throw new SemanticGenericError({message: error.message});
```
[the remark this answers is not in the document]
alenakhineika: Thx for reminding about it. I have added this type error validation everywhere when we execute JS. I will create tests for it in nest PR, when will handle validation for rest of constructors.

test/json/error/javascript/bson-constructors.json
```diff
@@ -2,9 +2,116 @@
...
+        "query": "Code('string', {x: '1'}, 3)",
+        "errorCode": "E_SEMANTIC_ARGUMENTCOUNTMISMATCH"
+      },
+      {
+        "description": "Code with not object scope",
+        "query": "new Code('string', 5)",
+        "errorCode": "E_SEMANTIC_TYPE"
+      }
+    ],
+    "ObjectId": [
```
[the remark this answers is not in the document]
alenakhineika: I didn't implemented validation for ObjectId in current PR. Will do it in next PR

codegeneration/javascript/Visitor.js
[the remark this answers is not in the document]
aherlihy: I don't think it is good practice to add functionality in one PR and then add tests to the next PR. You should either add tests for this in this PR, or if you really don't want to write tests here, you should remove this and add it to the next one. Or we can merge the PRs at the same time.

PR #67, opened 2018-05-18
fix object methods for c#

lrlna:
some of the changes made in this branch got overwritten by a rebase before this. Just opening this up to see if I can bring them back into master.


lrlna:
ok it doesn't actually get any better with the rebase; will have to fix in a new branch

PR #69, opened 2018-05-18
fix csharp bugs

lrlna:
I've started pulling all of the c# test results into an actual c# executable file (over in [c#-run-test](https://github.com/mongodb-js/bson-compilers/tree/c%23-run-test) branch), and I realized a bunch of stuff isn't actually compiling properly :O 

This PR fixes most things, but I am still waiting on some argument help over in #65.


codegeneration/csharp/Generator.js
```diff
@@ -47,6 +47,20 @@ module.exports = (superclass) => class ExtendedVisitor extends superclass {
...
+   *
+   * @param {NumberContextObject} ctx
+   *
+   * @returns {string} - (int)value
+   */
+  emitNumber(ctx) {
+    ctx.type = this.Types.Number;
+    const args = ctx.arguments().argumentList().singleExpression();
+    const expr = args[0].getText().replace(/['"]+/g, '');
+    return `(int)${expr}`;
```
lrlna: this number situation is a bit weird. to get to do `int` [declaration inline](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/int#conversions) you've got to do this thing ^

test/json/success/javascript/bson-constructors.json
```diff
@@ -48,23 +48,23 @@
         "javascript": "ObjectId()",
         "python": "ObjectId()",
         "java": "new ObjectId()",
-        "csharp": "new BsonObjectId()",
+        "csharp": "new ObjectId()",
```
lrlna: switching this over to `ObjectId()` since the compiler was warning to use this one instead of `BsonObjectId`

test/json/success/javascript/bson-constructors.json
```diff
@@ -152,23 +152,23 @@
         "javascript": "Double(3)",
         "python": "float(3)",
         "java": "new java.lang.Double(3)",
-        "csharp": "new BsonDouble(\"3\")",
+        "csharp": "new BsonDouble(3)",
```
lrlna: this needs a number! woooooops

test/json/success/javascript/bson-constructors.json
```diff
@@ -178,7 +178,7 @@
         "javascript": "new Long(-1, 2147483647)",
         "python": "Int64(9223372036854775807)",
         "java": "new java.lang.Long(\"9223372036854775807\")",
-        "csharp": "new BsonInt64.ToInt64(9223372036854775807)",
```
lrlna: this was not a valid thing; just needed to pass in the number to `BsonInt64` it seems

test/json/success/javascript/bson-methods.json
```diff
@@ -186,175 +186,175 @@
...
+        "csharp": "new BsonInt64(429496729601) < 0",
         "shell": "NumberLong(429496729601) < 0"
       },
       {
         "description": "isOdd",
         "javascript": "Long(1, 100).isOdd()",
         "python": "(Int64(429496729601) % 2) == 0",
         "java": "new java.lang.Long(\"429496729601\") % 2 == 0",
-        "csharp": "new BsonInt64.ToInt64(429496729601) % 2 == 0",
+        "csharp": "new BsonInt64(429496729601).ToInt64() % 2 == 0",
```
lrlna: there are a series of these changes where it doesn't compile unless i convert it to either `ToInt64` or to `ToInt32`, so now I am doing that.

And this was the error if you're curious:
```shell
Operator '+' cannot be applied to operands of type 'BsonInt64' and 'BsonInt64'
```

test/json/success/javascript/bson-methods.json
```diff
@@ -186,175 +186,175 @@
...
+        "csharp": "new BsonInt64(429496729601).ToInt32() << 10",
         "shell": "NumberLong(429496729601) << 10"
       },
       {
         "description": "shiftRight",
         "javascript": "Long(1, 100).shiftRight(10)",
         "python": "Int64(429496729601) >> 10",
         "java": "java.lang.Long.rotateRight(new java.lang.Long(\"429496729601\"), 10)",
-        "csharp": "new BsonInt64.ToInt64(429496729601) >> 10",
+        "csharp": "new BsonInt64(429496729601).ToInt32() >> 10",
```
lrlna: this one needs to be converted to `int32` since `10` is also an `int32`

test/json/success/javascript/bson-utils.json
```diff
@@ -7,22 +7,22 @@
...
+        "csharp": "new ObjectId(\"5ab901c29ee65f5c8550c5b9\")"
       },
       {
         "description": "createFromTime",
         "javascript": "ObjectId.createFromTime(1522075031642)",
         "python": "ObjectId.from_datetime(datetime.datetime.fromtimestamp(1522075031642/1000.0))",
         "java": "new ObjectId(new java.util.Date(new java.lang.Long(\"1522075031642\")))",
         "shell": "ObjectId.fromDate(new Date(1522075031642))",
-        "csharp": "new BsonObjectId.GenerateNewId(1522075031642)"
+        "csharp": "new ObjectId().GenerateNewId(Convert.ToInt32(1522075031642))"
```
lrlna: need to pass in an `int32` to this method.

there is still an issue with this in the compiler, but i'll leave it as this for now and fix it once i get clarification as to how to use this method.

and this is the current error if you're wondering:
```shell
Member 'ObjectId.GenerateNewId(int)' cannot be accessed with an instance reference; qualify it with a type name instead
```

test/json/success/javascript/bson-utils.json
```diff
@@ -115,55 +115,55 @@
...
-        "csharp": "new BsonInt64.ToInt64(\"5\")"
+        "csharp": "new BsonInt64(5)"
       },
       {
         "description": "fromString with radix",
         "javascript": "Long.fromString('5', 10)",
         "python": "Int64(int('5'))",
         "java": "java.lang.Long.parseLong(\"5\", 10)",
         "shell": "NumberLong('5')",
-        "csharp": "Int64.TryParse(\"5\", 10)"
```
lrlna: I can't figure out how to get `TryParse` to work since it's asking for an `out` method, so I am switching it over to `Convert.ToInt64` which also takes a `radix`

test/json/success/javascript/language-types.json
```diff
@@ -254,95 +254,95 @@
...
-        "csharp": "new Regex(@\"(?im)ab+c\")",
+        "csharp": "new Regex(\"(?im)ab+c\")",
         "shell": "new RegExp('ab+c', 'im')"
       },
       {
         "description": "regex object with ig flags as args",
         "javascript": "new RegExp('ab+c', 'ig')",
         "python": "re.compile(r\"ab+c(?is)\")",
         "java": "Pattern.compile(\"ab+c(?i)\")",
-        "csharp": "new Regex(@\"(?i)ab+c\")",
```
lrlna: getting rid of all the `@` signs since they interfere with when you want to escape characters.

test/json/success/javascript/bson-utils.json
aherlihy: That is a very confusing error message.

test/json/success/javascript/bson-utils.json
```diff
@@ -173,7 +173,7 @@
         "python": "Decimal128('5')",
         "java": "Decimal128.parse(\"5\")",
         "shell": "NumberDecimal('5')",
-        "csharp": "new Decimal128.TryParse(\"5\")"
+        "csharp": "new Decimal128(5)"
```
aherlihy: I think we should try not to convert argument types, if we can avoid it, in case there's some kind of overflow error or if the string argument isn't valid. Maybe we can ask craig what the most normal way to get a Decimal128 from a string is.

test/json/success/javascript/bson-methods.json
lrlna: there might be! I asked already as to the best way of doing this. I'd understand if the two types doing the math were different, but they are both `BsonInt64`, so i am confused hah o/

test/json/success/javascript/bson-utils.json
lrlna: yea haha asked also about this one; it's beyond my understanding

codegeneration/csharp/Generator.js
```diff
@@ -79,7 +93,7 @@ module.exports = (superclass) => class ExtendedVisitor extends superclass {
    * @param {BSONRegExpConstructorObject} ctx - expects two strings as
```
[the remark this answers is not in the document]
lrlna: ohhhhhh dang

test/json/success/javascript/bson-utils.json
lrlna: i'll ask about this, yea. It seems weird to coerce, which is basically what i am doing o/

lrlna:
i am going to merge this, and then make changes once i get more information on some of the proper ways of doing stuff

lrlna:
(and mainly merging right now because otherwise i'll be stuck in rebase land™ for like a day if i wait on the other branch)

PR #76, opened 2018-05-22
[WIP] Verify java tests

aherlihy:
Add back the java tests so that Jeff can see/comment on them directly.

### DIFF
--- a/test/json/success/javascript/bson-constructors.json
+++ b/test/json/success/javascript/bson-constructors.json
@@ -5,7 +5,7 @@
         "description": "new Code with string arg",
         "javascript": "new Code('some code')",
         "python": "Code('some code')",
-        "java": "",
+        "java": "new Code(\"some code\")",
         "csharp": "new BsonJavaScript(@\"some code\")",
         "shell": "Code('some code')"
       },
@@ -13,7 +13,7 @@
         "description": "Code with string code",
         "javascript": "Code('some code')",
         "python": "Code('some code')",
-        "java": "",
+        "java": "new Code(\"some code\")",
         "csharp": "new BsonJavaScript(@\"some code\")",
         "shell": "Code('some code')"
       },
@@ -21,7 +21,7 @@
         "description": "Code with string code and object scope",
         "javascript": "Code('string', {x: '1'})",
         "python": "Code('string', {'x': '1'})",
-        "java": "",
+        "java": "new CodeWithScope(\"string\", new Document().append(\"x\", \"1\"))",
         "csharp": "new BsonJavaScriptWithScope(@\"string\", new BsonDocument(\"x\", \"1\"))",
         "shell": "Code('string', {x: '1'})"
       },
@@ -29,7 +29,7 @@
         "description": "Code with function code",
         "javascript": "Code(function(test) { console.log(test); })",
         "python": "Code('function(test){console.log(test);}')",
-        "java": "",
+        "java": "new Code(\"function(test){console.log(test);}\")",
         "csharp": "new BsonJavaScript(@\"function(test){console.log(test);}\")",
         "shell": "Code('function(test){console.log(test);}')"
       },
@@ -37,7 +37,7 @@
         "description": "Code with where code",
         "javascript": "new Code('this.a > i', { i: '1' })",
         "python": "Code('this.a > i', {'i': '1'})",
-        "java": "",
+        "java": "new CodeWithScope(\"this.a > i\", new Document().append(\"i\", \"1\"))",
         "csharp": "new BsonJavaScriptWithScope(@\"this.a > i\", new BsonDocument(\"i\", \"1\"))",
         "shell": "Code('this.a > i', {i: '1'})"
       }
@@ -47,7 +47,7 @@
         "description": "ObjectId with no arg",
         "javascript": "ObjectId()",
         "python": "ObjectId()",
-        "java": "",
+        "java": "new ObjectId()",
         "csharp": "new ObjectId()",
         "shell": "ObjectId()"
       },
@@ -55,7 +55,7 @@
         "description": "new ObjectId with no arg",
         "javascript": "new ObjectId()",
         "python": "ObjectId()",
-        "java": "",
+        "java": "new ObjectId()",
         "csharp": "new ObjectId()",
         "shell": "new ObjectId()"
       },
@@ -63,7 +63,7 @@
         "description": "ObjectId with hex string arg",
         "javascript": "ObjectId('5a7382114ec1f67ae445f778')",
         "python": "ObjectId('5a7382114ec1f67ae445f778')",
-        "java": "",
+        "java": "new ObjectId(\"5a7382114ec1f67ae445f778\")",
         "csharp": "new ObjectId(\"5a7382114ec1f67ae445f778\")",
         "shell": "new ObjectId('5a7382114ec1f67ae445f778')"
       }
@@ -73,7 +73,7 @@
         "description": "Binary with ascii buffer arg",
         "javascript": "Binary(Buffer.from('1234'))",
         "python": "Binary(b'1234')",
-        "java": "",
+        "java": "new Binary(\"1234\".getBytes(\"UTF-8\"))",
         "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"))",
         "shell": "new BinData(4, '1234')"
       },
@@ -81,7 +81,7 @@
         "description": "new Binary with (ascii buffer, UUID subtype enum) args",
         "javascript": "new Binary(Buffer.from(\"1234\"), Binary.SUBTYPE_UUID)",
         "python": "Binary(b'1234', bson.binary.UUID_SUBTYPE)",
-        "java": "",
+        "java": "new Binary(org.bson.BsonBinarySubType.UUID_STANDARD, \"1234\".getBytes(\"UTF-8\"))",
         "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"), BsonBinarySubType.UuidStandard)",
         "shell": "new BinData(4, '1234')"
       },
@@ -89,7 +89,7 @@
         "description": "Binary with (ascii buffer, number) args",
         "javascript": "Binary(Buffer.from('1234'), '1')",
         "python": "Binary(b'1234', bson.binary.FUNCTION_SUBTYPE)",
-        "java": "",
+        "java": "new Binary(org.bson.BsonBinarySubType.FUNCTION, \"1234\".getBytes(\"UTF-8\"))",
         "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"), BsonBinarySubType.Function)",
         "shell": "new BinData(1, '1234')"
       },
@@ -97,7 +97,7 @@
         "description": "Binary with string args",
         "javascript": "Binary('1234')",
         "python": "Binary(b'1234')",
-        "java": "",
+        "java": "new Binary(\"1234\".getBytes(\"UTF-8\"))",
         "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"))",
         "shell": "new BinData(1, '1234')"
       },
@@ -105,7 +105,7 @@
         "description": "Binary with (string, string) args",
         "javascript": "Binary('1234', '1')",
         "python": "Binary(b'1234', bson.binary.FUNCTION_SUBTYPE)",
-        "java": "",
+        "java": "new Binary(org.bson.BsonBinarySubType.FUNCTION, \"1234\".getBytes(\"UTF-8\"))",
         "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"), BsonBinarySubType.Function)",
         "shell": "new BinData(4, '1234')"
       }
@@ -115,7 +115,7 @@
         "description": "new DBRef with (string, ObjectID) args",
         "javascript": "new DBRef('coll', new ObjectId())",
         "python": "DBRef('coll', ObjectId())",
-        "java": "",
+        "java": "new DBRef(\"coll\", new ObjectId())",
         "csharp": "new MongoDBRef(\"coll\", new ObjectId())",
         "shell": "new DBRef('coll', new ObjectId())"
       },
@@ -123,7 +123,7 @@
         "description": "new DBRef with (string, ObjectId, string) args",
         "javascript": "new DBRef('coll', ObjectId(), 'db')",
         "python": "DBRef('coll', ObjectId(), 'db')",
-        "java": "",
+        "java": "new DBRef(\"db\", \"coll\", new ObjectId())",
         "csharp": "new MongoDBRef(\"coll\", new ObjectId(), \"db\")",
         "shell": "new DBRef('coll', ObjectId(), 'db')"
       }
@@ -133,7 +133,7 @@
         "description": "Int32 with number arg",
         "javascript": "Int32(3)",
         "python": "int(3)",
-        "java": "",
+        "java": "new java.lang.Integer(3)",
         "csharp": "Convert.ToInt32(3)",
         "shell": "NumberInt(3)"
       },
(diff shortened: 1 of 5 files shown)


test/json/success/javascript/bson-utils.json
```diff
@@ -5,31 +5,31 @@
...
+        "java": "new ObjectId(\"5ab901c29ee65f5c8550c5b9\")",
         "shell": "ObjectId('5ab901c29ee65f5c8550c5b9')",
         "csharp": "new ObjectId(\"5ab901c29ee65f5c8550c5b9\")"
       },
       {
         "description": "createFromTime",
         "javascript": "ObjectId.createFromTime(1522075031642)",
         "python": "ObjectId.from_datetime(datetime.datetime.fromtimestamp(1522075031642/1000.0))",
-        "java": "",
+        "java": "new ObjectId(new java.util.Date(new java.lang.Long(\"1522075031642\")))",
```
jyemin: Just use a long literal

test/json/success/javascript/bson-utils.json
```diff
@@ -39,55 +39,55 @@
...
+        "java": "org.bson.BsonBinarySubType.FUNCTION",
         "shell": "1",
         "csharp": "BsonBinarySubType.Function"
       },
       {
         "description": "subtype byte array",
         "javascript": "Binary.SUBTYPE_BYTE_ARRAY",
         "python": "bson.binary.OLD_BINARY_SUBTYPE",
-        "java": "",
+        "java": "org.bson.BsonBinarySubType.OLD_BINARY",
```
jyemin: Should be BINARY not OLD_BINARY

test/json/success/javascript/bson-utils.json
```diff
@@ -39,55 +39,55 @@
...
         "shell": "2",
         "csharp": "BsonBinarySubType.OldBinary"
       },
       {
         "description": "subtype uuid old",
         "javascript": "Binary.SUBTYPE_UUID_OLD",
         "python": "bson.binary.OLD_UUID_SUBTYPE",
-        "java": "",
+        "java": "org.bson.BsonBinarySubType.UUID_LEGACY",
```
jyemin: We have to be very careful about UUIDs because of all the legacy behavior from various drivers.  What is this being used for?

test/json/success/javascript/bson-utils.json
```diff
@@ -97,71 +97,71 @@
...
+        "java": "java.lang.Long.MIN_VALUE",
         "shell": "math.min()",
         "csharp": "Int64.MinValue"
       },
       {
         "description": "ZERO",
         "javascript": "Long.ZERO",
         "python": "Int64(0)",
-        "java": "",
+        "java": "new java.lang.Long(0)",
```
jyemin: Don't box it.  Just use `0L`

test/json/success/javascript/bson-utils.json
```diff
@@ -97,71 +97,71 @@
...
+        "java": "new java.lang.Long(-1)",
         "shell": "-1",
         "csharp": "new BsonInt64(-1)"
       },
       {
         "description": "fromInt",
         "javascript": "Long.fromInt(5)",
         "python": "Int64(5)",
-        "java": "",
+        "java": "new java.lang.Long(\"5\")",
```
jyemin: Just cast: `(long) 5`  or use a long literal `5L`.  Depends on what the context is.

test/json/success/javascript/bson-utils.json
```diff
@@ -97,71 +97,71 @@
...
+        "java": "java.lang.Long.parseLong(\"5\", 10)",
         "shell": "NumberLong('5')",
         "csharp": "Convert.ToInt64(\"5\", 10)"
       },
       {
         "description": "fromBits",
         "javascript": "Long.fromBits(-1, 2147483647)",
         "python": "Int64(9223372036854775807)",
-        "java": "",
+        "java": "new java.lang.Long(\"9223372036854775807\")",
```
jyemin: What is the semantics of fromBits?

test/json/success/javascript/language-types.json
```diff
@@ -5,15 +5,15 @@
         "description": "simple addition",
         "javascript": "2+5",
         "python": "2+5",
-        "java": "",
+        "java": "new java.lang.Long(\"2\")+new java.lang.Long(\"5\")",
```
jyemin: Why not `2 + 5`?

test/json/success/javascript/language-types.json
```diff
@@ -41,119 +41,119 @@
         "description": "Integer 2",
         "javascript": "2",
         "python": "2",
-        "java": "",
+        "java": "new java.lang.Long(\"2\")",
```
jyemin: Integer and Long are different in Java.  Long is 64 bits and Integer is 32 bits.  They map to BSON Int64 and BSON Int32, respectively.  So this won't result in an Int32 in MongoDB, if that's what you're after.

test/json/success/javascript/language-types.json
```diff
@@ -41,119 +41,119 @@
...
+        "java": "new java.lang.Long(\"429496729601\")",
         "csharp": "429496729601",
         "shell": "429496729601"
       },
       {
         "description": "Decimal 2.001",
         "javascript": "2.001",
         "python": "2.001",
-        "java": "",
+        "java": "new java.lang.Double(2.001)",
```
jyemin: Don't box it.

test/json/success/javascript/language-types.json
```diff
@@ -41,119 +41,119 @@
...
         "csharp": "2.001",
         "shell": "2.001"
       },
       {
         "description": "Hex number caps",
         "javascript": "0X123ABC",
         "python": "0X123ABC",
-        "java": "",
+        "java": "0X123ABC",
```
jyemin: In what context will this be used?

test/json/success/javascript/language-types.json
```diff
@@ -41,119 +41,119 @@
...
+        "java": "null",
         "csharp": "BsonNull.Value",
         "shell": "null"
       },
       {
         "description": "undefined",
         "javascript": "undefined",
         "python": "None",
-        "java": "",
+        "java": "null",
```
jyemin: You can use `new BsonUndefined()`

test/json/success/javascript/bson-constructors.json
```diff
@@ -73,39 +73,39 @@
...
+        "java": "new Binary(\"1234\".getBytes(\"UTF-8\"))",
         "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"1234\"))",
         "shell": "new BinData(4, '1234')"
       },
       {
         "description": "new Binary with (ascii buffer, UUID subtype enum) args",
         "javascript": "new Binary(Buffer.from(\"1234\"), Binary.SUBTYPE_UUID)",
         "python": "Binary(b'1234', bson.binary.UUID_SUBTYPE)",
-        "java": "",
+        "java": "new Binary(org.bson.BsonBinarySubType.UUID_STANDARD, \"1234\".getBytes(\"UTF-8\"))",
```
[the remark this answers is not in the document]
aherlihy: So it would be better to use `UUID_STANDARD` directly?

We are probably going to provide an import string along with the generated code, so we could import the subtype classes directly.

test/json/success/javascript/bson-methods.json
```diff
@@ -66,31 +66,31 @@
...
+        "java": "new Binary(\"a string\".getBytes(\"UTF-8\")).getData()",
         "csharp": "new BsonBinaryData(System.Text.Encoding.ASCII.GetBytes(\"a string\")).ToString()",
         "shell": "BinData(1, '0001').hex()"
       },
       {
         "description": "toString",
         "javascript": "Binary(Buffer.from('a string')).toString()",
         "python": "str(Binary(b'a string'))",
-        "java": "",
+        "java": "new Binary(\"a string\".getBytes(\"UTF-8\")).toString()",
```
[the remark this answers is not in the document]
aherlihy: All these tests are from javascript --> other languages. What javascript does, it seems, is create a utf-8 string [node driver Binary.toString](https://github.com/mongodb/js-bson/blob/1.0-branch/lib/bson/binary.js#L268) and the [node Buffer toString](https://www.w3schools.com/nodejs/met_buffer_tostring.asp) defaults to utf-8.

test/json/success/javascript/bson-methods.json
```diff
@@ -134,7 +134,7 @@
         "description": "valueOf",
         "javascript": "Int32(3).valueOf()",
         "python": "int(3)",
-        "java": "",
+        "java": "new java.lang.Integer(3).intValue()",
```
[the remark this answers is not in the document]
aherlihy: Makes sense, I can change it to just an integer. 

(It was there because in the languages where Int32 is needed to force the integer size, there also needs to be a way to get the numeric value out of the object. This was just a direct translation from the javascript, but you're right it doesn't make sense for a lang where the default int size is 32).

test/json/success/javascript/language-types.json
```diff
@@ -163,87 +163,87 @@
...
+        "java": "new java.util.Date(new java.lang.Long(\"819170640000\"))",
         "csharp": "new DateTime(1995, 12, 17, 3, 24, 0)",
         "shell": "new Date(1995, 11, 17, 3, 24, 0)"
       },
       {
         "description": "Date string with ISO String",
         "javascript": "Date('December 17, 1995 03:24:00Z')",
         "python": "datetime.datetime(1995, 12, 17, 3, 24, 0, tzinfo=datetime.timezone.utc).strftime('%a %b %d %Y %H:%M:%S %Z')",
-        "java": "",
+        "java": "new java.util.Date(new java.lang.Long(\"819170640000\")).toString()",
```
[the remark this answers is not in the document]
aherlihy: `Date(...)` in JS generates a string with the format 'dow mon dd yyyy hh:mm:ss zzz'. I missed that the year was in the wrong place. Any issue with using: SimpleDateFormat (i.e. `SimpleDateFormat("EEE MMMMM dd yyyy HH:mm:ss").format(new java.util.Date(819170640000L))`)?

test/json/success/javascript/bson-utils.json
aherlihy: Just trying for equivalence between drivers subtype flags. Does the javascript SUBTYPE_UUID_OLD (which is the shell type \x03) match the java drivers UUID_LEGACY?

test/json/success/javascript/bson-utils.json
aherlihy: It is the same as the timestamp constructor, it takes a low bit and a high bit. Since there's no equivalent function (is there?) in java for Long, I just calculate what the long would be (by executing it in javascript) and construct a long from that. I can make it into `9223372036854775807L` no problem.

test/json/success/javascript/language-types.json
aherlihy: Numeric literals in javascript (and most languages I think, except java) are 64-bit, so they should get converted to Long (or Double, if it has a decimal) so not to have a compilation error if the number is too large for 32-bits. Can replace it with `2L + 5L`, but since there's no telling if the value is too big for 32-bits, I can't just compile it as 2+5.

test/json/success/javascript/language-types.json
aherlihy: Since it's from javascript originally, it's a 64-bit integer getting compiled.

test/json/success/javascript/language-types.json
aherlihy: I guess if a user wanted to compile javascript code that included a hex value? I haven't been spending too much time trying to guess what's useful (within reason), I'm trying to get parity across languages. We could remove support for hex literals in javascript, and not bother converting them, but unless it's incorrect there's not much harm in keeping them even if I don't anticipate that lots of people are going to be compiling hex values with literals.

test/json/success/javascript/bson-methods.json
```diff
@@ -5,23 +5,23 @@
...
+        "java": "new Code(\"test code\").getCode()",
         "csharp": "new BsonJavaScript(@\"test code\").Code",
         "shell": "Code('test code').code"
       },
       {
         "description": "Code toString",
         "javascript": "Code('test code').toString()",
         "python": "repr(Code('test code'))",
-        "java": "",
+        "java": "new Code(\"test code\").toString()",
```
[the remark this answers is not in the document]
aherlihy: I might delete this method, it's not a very useful conversion because even in javascript, toString just produces `'[object Object]'`

test/json/success/javascript/bson-constructors.json
```diff
@@ -133,15 +133,15 @@
         "description": "Int32 with number arg",
         "javascript": "Int32(3)",
         "python": "int(3)",
-        "java": "",
+        "java": "new java.lang.Integer(3)",
```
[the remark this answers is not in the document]
aherlihy: And when there's a string argument, should I use `new java.lang.Integer("3")` or is it better to use `Integer.parseInt("3")`? (same for `Double.parseDouble("3.001")`?)

test/json/success/javascript/bson-constructors.json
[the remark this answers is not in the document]
jyemin: I don't have enough context to understand how this would be used, but this line of code makes no sense.  The UUID Standard subtype has to be of the correct format (see https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/codecs/UuidCodec.java) so it makes no sense to convert a string to UTF8 and store it as UUID Standard

test/json/success/javascript/language-types.json
jyemin: Ack.  I think this is ok.

test/json/success/javascript/bson-constructors.json
```diff
@@ -73,39 +73,39 @@
         "description": "Binary with ascii buffer arg",
         "javascript": "Binary(Buffer.from('1234'))",
         "python": "Binary(b'1234')",
-        "java": "",
+        "java": "new Binary(\"1234\".getBytes(\"UTF-8\"))",
```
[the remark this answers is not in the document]
aherlihy: This is for testing if we can take arbitrary Binary object in javascript and convert it to a Binary object in java. It's "1234" because it's easy to type, but it could be anything (of type String, Buffer, Uint8Array or Array). We don't care what the contents of that binary object is, we just want it to end up the same type in java.

We are doing the conversion by executing the javascript input (since we don't know/care what the actual argument is to Binary, as long as it's the correct type):

    b = Binary(Buffer.from('010101'))
Then extracting the value of the Binary object as a string

    str = b.value
Then using that string to construct a java object (escaping any double quotes)

    return `new Binary("${str}".getBytes())`

Are there cases in which this wouldn't work?

test/json/success/javascript/language-types.json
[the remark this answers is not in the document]
jyemin: SimpleDateFormat is the way forward!

test/json/success/javascript/bson-constructors.json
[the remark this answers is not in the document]
aherlihy: Is it invalid java, or does the test case not make sense? If a user creates a binary object, and wants to set the subtype as UUID, we want to pass along the data so that an equivalent UUID Binary data object in java gets generated. We are testing that we can rearrange the code so that the result is equivalent, not testing if we can store "1234" as a UUID Binary data object. I'm trying to figure out if the translation is invalid, or if it's just a silly test case that doesn't make sense to actually run.

test/json/success/javascript/language-types.json
```diff
@@ -5,15 +5,15 @@
...
+        "java": "2L+5L",
         "csharp": "2+5",
         "shell": "2+5"
       },
       {
         "description": "multiplication and division",
         "javascript": "2+(4*36)/3",
         "python": "2+(4*36)/3",
-        "java": "new java.lang.Long(\"2\")+(new java.lang.Long(\"4\")*new java.lang.Long(\"36\"))/new java.lang.Long(\"3\")",
+        "java": "2L+(4L*36L)/3L",
```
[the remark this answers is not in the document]
aherlihy: Since all numbers in javascript are 64-bit, and we don't check if the number is too large for Int32, we always generate javascript integers as java longs.

aherlihy:
Closing because got all the feedback needed. Will eventually open another PR with the Binary data changes + the java builders code.

PR #82, opened 2018-05-28
Idiomatic Java Syntax

aherlihy:
- Use `BsonUndefined` instead of null
- `java.lang` is always imported, so don't need to use the full name.
- Use `SimpleDateFormat` instead of `toString()`
- Use `toHexString` instead of `toString` for ObjectId
- Use `BINARY` instead of `OLD_BINARY` for Binary subtype.


aherlihy:
No unfortunately, there are a few big things (like builders) and a few
small things (like writing better tests) that are still to go.

PR #83, opened 2018-05-28
Idiomatic Java Numbers

aherlihy:
Adds a `processNumber` method so that the templates get passed an extra argument that indicates the type of the argument. This is useful because now we can translate:

Double(10) --> 10d
Double("10") --> Double.parseDouble("10")
Double(0x10) --> new Double(0x10)

Without adding any emit methods.

All the tests are passing except the C# tests because I didn't want to modify code that Irina's actively working on. @lrlna - If this update looks good to you then we can convert the C# numbers over to this. It should be really straightforward: remove number emit methods, and in the templates, check if the `type` argument is `_string`: if so, generate the parsing constructor instead of the normal one.


codegeneration/javascript/Visitor.js
```diff
@@ -357,6 +360,21 @@ class Visitor extends ECMAScriptVisitor {
     return res;
   }
 
+  getActualTyped(actual) {
+    if (actual.type === undefined && actual.actualType === undefined) {
+      while (actual.singleExpression()) {
+        actual = actual.singleExpression();
+        if (actual.type !== undefined || actual.actualType !== undefined) {
```
lrlna: i think `actual.type` and `actual.actualType` *might* get a bit confusing. In fact, I am not sure what the difference is meant to be since we assign `actual.type` to `actual.actualType` in `getTyped()` func below.

aherlihy: haha yes, definitely poor wording. 

The difference between type and actualType comes in when we're casting argument values. Since the input `"5"` gets generated as `5d` because all numbers are doubles in javascript, we have to be careful we're not applying that to arguments unnecessarily.

So if we didn't do any casting, then the input `"Double(5)"` would get generated as `5dd`, because the argument becomes `5d` and then the `Double` call adds _another_ `d` to the output. We can get around this by specifying in the Types that Double takes in an Integer and integers aren't converted. So long story short, the "actual" type of the argument would be "Double", but the "type" of the argument is "Integer" because we've casted it.

For argument "actual", that is just indicating that it's the input from the user. It's called from castType which is given an "actual" and an "expected", so it's "actual" in a completely different sense. My bad, I will change the names so that it's not so confusing.

PR #85, opened 2018-05-30
COMPASS-2804 write error tests for python

alenakhineika:
Error tests for python

### DIFF
--- a/codegeneration/javascript/Visitor.js
+++ b/codegeneration/javascript/Visitor.js
@@ -513,8 +513,26 @@ class Visitor extends ECMAScriptVisitor {
    */
   processRegExp(ctx) {
     ctx.type = this.Types._regex;
+    const text = ctx.getText();
     let pattern;
     let flags;
+
+    if (text.startsWith('RegExp')) {
+      const argList = ctx.arguments().argumentList();
+
+      if (
+        !argList ||
+        !(
+          argList.singleExpression().length === 1 ||
+          argList.singleExpression().length === 2
+        )
+      ) {
+        throw new SemanticArgumentCountMismatchError({
+          message: 'RegExp requires one or two arguments'
+        });
+      }
+    }
+
     try {
       const regexobj = this.executeJavascript(ctx.getText());
       pattern = regexobj.source;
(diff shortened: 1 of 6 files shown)


test/json/success/javascript/bson-methods.json
```diff
@@ -279,7 +279,7 @@
       },
       {
         "description": "negate",
-        "javascript": "Long(1, 100).negate())",
```
lrlna: holly cow how was is passing!

alenakhineika: I suppose it looks at `negate` and assumes that argument should go next. Since `negate` doesn't have arguments and it is valid scenario, it skips everything what goes next. It also works:

codegeneration/javascript/Visitor.js
```diff
@@ -513,8 +513,26 @@ class Visitor extends ECMAScriptVisitor {
    */
   processRegExp(ctx) {
     ctx.type = this.Types._regex;
+    const text = ctx.getText();
     let pattern;
     let flags;
+
+    if (text.startsWith('RegExp')) {
```
aherlihy: Why can't we use checkArguments here?

codegeneration/javascript/Visitor.js
```diff
@@ -728,6 +754,14 @@ class Visitor extends ECMAScriptVisitor {
     ctx.type = this.Types._string;
     const long = ctx.singleExpression().singleExpression();
     let longstr;
+    const argList = ctx.arguments().argumentList();
```
aherlihy: Same question here

test/json/error/javascript/bson-constructors.json
```diff
@@ -158,10 +173,110 @@
...
       {
-        "description": "BSONRegExp with not a string flaf",
+        "description": "BSONRegExp with not a string flag",
         "query": "new BSONRegExp('^[a-z0-9_-]{3,16}$', 123)",
         "errorCode": "E_SEMANTIC_TYPE"
       }
+    ],
+    "Decimal128": [
+      {
+        "description": "Decimal128 without argumnets",
```
aherlihy: arguments is misspelled in all these tests

PR #87, opened 2018-05-30
Add file for generating target language files

aherlihy:
Generate files that run on their own so that we can make sure the syntax in our tests compile + run. 

TODO: Should maybe add instructions on how to get them to run?

### DIFF
--- a/make-test-file.js
+++ b/make-test-file.js
@@ -10,24 +10,16 @@ const javaFileTemplate = (code) => {
   return `
 package com.example.test;
 
+import com.mongodb.DBRef;
 import org.bson.BsonBinarySubType;
 import org.bson.BsonRegularExpression;
 import org.bson.Document;
 import org.bson.types.*;
-import org.bson.codecs.PatternCodec;
+import org.bson.BsonUndefined;
 
-import com.mongodb.*;
-
-
-import javax.print.Doc;
+import java.text.SimpleDateFormat;
 import java.util.Arrays;
-import java.util.List;
-import java.util.Set;
 import java.util.regex.Pattern;
-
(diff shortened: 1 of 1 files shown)


make-test-file.js
```diff
@@ -0,0 +1,144 @@
...
+      return str0 + Object.keys(tests).reduce(
+        (str, key) => {
+          return str + templates[output].doc(file.replace(/-/g, '').slice(0, -5) + key, tests[key].reduce(
+            (str2, test) => {
+              return str2 + templates[output].line(test.description, test[output]);
+            }, ''));
+        }, '');
+    }, ''));
+};
+console.log(process.argv);
```
lrlna: should remove this since if you want to do `node make-test-file > Program.cs` the file that ends up in the `Program.cs` is not executable

make-test-file.js
```diff
@@ -0,0 +1,144 @@
...
+if (process.argv.length !== 3) {
```
lrlna: this is more of a stylistic comment: i would move the argument parsing on top. The usual pattern with node's CLI-type of applications is to have argument processing first before you get into the rest of the stuff. it also makes it clear as to what the file does

make-test-file.js
aherlihy: Oops!

make-test-file.js
aherlihy: Cool! I updated it so that the argument parsing is on top + added the shebang. I didn't add minimist yet since it doesn't seem like we need anything else from the arg parsing at the moment.

aherlihy:
@lrlna This PR should be merged after #88 so that it can use the import strings directly.

lrlna:
squashing and merging as well! woot woot!

PR #88, opened 2018-05-30
Add import string as export

aherlihy:
Each language exports a string including the required imports now. Can be used in the UI to indicate which packages are being used.

Currently, doesn't do anything "smart", and just lists all the packages that can be used. In the future we could add detection for classes that are getting used and only show imports for those classes, but for now we import everything.

TODO: make sure we're importing in the most 'idiomatic' way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,18 +19,20 @@ const string =`
 
 const compiledString = compiler[input][output](string);
 console.log(compiledCode)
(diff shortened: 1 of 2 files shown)


lrlna:
i am going to squash and merge this one! it looks grrr888

PR #90, opened 2018-05-31
Java Numeric Bugfix

aherlihy:
This PR has two bug fixes that I found when I ran the `make-test-file.js` for java.

1. Apparently you can't create numeric types given certain other numeric types, so for example `new Integer(1d)` is invalid, and instead it must be `(int) 1d`. The tests are templates are updated to reflect this now.

2. We were not casting octal arguments correctly for java, because `new Long(0o5)` is not valid java so we need to apply the octal template before the casted-to-type template, so we can do `new Long(05)`.


symbols/java/templates.yaml
```diff
@@ -596,7 +596,7 @@ Templates:
                 }
                 return arg;
             }
-            return `new Integer(${arg})`;
+            return `(int) ${arg}`;
```
lrlna: this syntax is soooo weird to me ahha c#'s got the same situation with the `int`

PR #94, opened 2018-06-04
add error documentation

lrlna:
I am not 100% that I got everything right in the pull request, so I would really love it if y'all look over this. @aherlihy @alenakhineika

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,57 +49,58 @@ Any compiler errors that occur will be thrown. To catch them, wrap the
 `compiler` in a `try/catch` block.
 - __error.message:__ Message `bson-compilers` will send back letting you know
   the compiler error.
-- __error.payload:__ The usual payload that comes from the error message.
 - __error.stack:__ The usual error stacktrace.
 - __error.code:__ Error code that `bson-compilers` adds to the error object to
   help you distinguish error types.
+- __error.line:__ If it is a syntax error, will have the line.
+- __error.column:__ If it is a syntax error, will have the column.
+- __error.symbol:__ If it is a syntax error, will have the symbol associated with the error.
 
-### error.code
-There are a few errorCodes `bson-compilers` sends back to help you figure out
-what went wrong during compilation:
+### Errors
+There are a few different error classes thrown by `bson-compilers`, each with 
+their own error code:
 
-##### E_SEMANTIC_ARGUMENTCOUNTMISMATCH
-This will occur when you're using a method with a wrong number of arguments.
+#### BsonCompilersArgumentError
+###### Code: E_BSONCOMPILERS_ARGUMENT
+This will occur when you're using a method with a wrong number of arguments, or
+the arguments are of the wrong type.
 For example, `ObjectId().equals()` requires one argument and it will throw if
 anything other than one argument is provided:
 
 ```javascript
-// ✘: this will throw a E_SEMANTIC_ARUGMENTCOUNTMISMATCH
+// ✘: this will throw a BsonCompilersArgumentError.
 ObjectId().equals(ObjectId(), ObjectId());
 
 // ✔: this won't throw
 ObjectId().equals(ObjectId());
 ```
 
-##### E_SEMANTIC_ATTRIBUTE
+```javascript
+// ✘: this will throw a BsonCompilersArgumentError.
+ObjectId({});
+
+// ✔: this won't throw
+ObjectId();
+```
+
+#### BsonCompilersAttributeError
+###### Code: E_BSONCOMPILERS_ATTRIBUTE
 Will be thrown if an invalid method or property is used on a BSON object. For
 example, since `new DBRef()` doesn't have a method `.foo()`, compiler will
 throw:
 
 ```javascript
-// ✘: method foo doesn't exist, so this will throw a E_SEMANTIC_ATTRIBUTE error 
+// ✘: method foo doesn't exist, so this will throw a BsonCompilersAttributeError .
 new DBRef('newCollection', new ObjectId()).foo()
 
 // ✔: this won't throw, since .toString() method exists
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,31 +15,128 @@ const input = 'javascript';
...
+new DBRef('newCollection', new ObjectId()).foo()
+
+// ✔: this won't throw, since .toString() method exists
+new DBRef('newCollection', new ObjectId()).toString(10)
+```
+
+##### E_SEMANTIC_GENERIC
+
+Semantic Generic error will be thrown if an unsupported argument is provided.
+For example when using a `RegExp()` an unsupported flag is given:
```
alenakhineika: E_SEMANTIC_GENERIC is used for any semantic error, which is not covered by rest of semantic errors. Maybe in the future it would be nice to expand a list of possible errors.

lrlna: ok cool! i was a bit confused by this type! that makes sense, thanks!

README.md
```diff
@@ -15,31 +15,128 @@ const input = 'javascript';
...
 ## API
 ### compiledString = compiler\[inputLang\]\[outputLang\](codeString)
 Output a compiled string given input and output languages.
-- __inputLang:__ Input language of the code string. `shell` and `javascript` are currently supported.
-- __outputLang:__ The language you would like the output to be. `java`, `python` and `csharp` are currently supported.
-- __codeString:__ The code string you would like to be compiled to your selected output language.
+- __inputLang:__ Input language of the code string. `shell` and `javascript`
+  are currently supported.
+- __outputLang:__ The language you would like the output to be. `java`,
+  `python` and `csharp` are currently supported.
```
alenakhineika: Also shell

lrlna: oh good point! that's from an outdated version from a couple of weeks ago

README.md
```diff
@@ -15,31 +15,128 @@ const input = 'javascript';
...
+number will throw an error:
+
+```javascript
+// ✘: the 'high' bit is not a number, so will throw E_SEMANTIC_TYPE
+Timestamp(10, {})
+
+// ✔: both params provided are numbers, will not throw
+Timestamp(10, 100)
+```
+
```
alenakhineika: We also have SemanticReferenceError

alenakhineika:
We also have SemanticReferenceError

lrlna:
@alenakhineika when does the Reference Error get thrown? I couldn't find it in the tests I don't think!

alenakhineika:
There is no such kind of error in tests, but it is being thrown in visitor.js files. I think we need include it also in read me, because read me is not only about tests. And also would be nice to have tests for this case. Not sure how to emulate a query to throw this error. Maybe we can ask Anna, she might know.

lrlna:
oh yea totally should include it in the readme if it gets thrown to the end user! I'll double check in the `visitor.js` file

README.md
aherlihy: And JS, I added that a few weeks ago

aherlihy:
@lrlna Exactly, SemanticReferenceError just a ReferenceError or a "symbol undefined" error.

We could give the exceptions more human-readable names. The only import thing to keep is the distinction between parsing errors and runtime (aka semantic) exceptions - because we don't throw parsing errors (ANTLR does) and those errors are fatal but our errors don't have to be.

At the start, I think I was wrong to suggest we tack on "semantic" to the error name. I now realized it's not as well-understood a word as I thought.  One naming pattern I like is in Python, when exceptions have simple names but inherit from classes so you can distinguish what type you have. Not sure we need to implement error inheritance now, but the names look like:

`E_SYNTAX_GENERIC` --> `SyntaxError`
`E_SEMANTIC_GENERIC`, --> `RuntimeError`
`E_SEMANTIC_TYPE` --> `TypeError`
`E_SEMANTIC_REFERENCE` --> `NameError`
`E_SEMANTIC_ARGUMENTCOUNTMISMATCH` --> In python this is a `TypeError`, but we can define our own `ArgumentCountError`
`E_SEMANTIC_ATTRIBUTE` --> `AttributeError`

What do you think? If it's just a rename, it'll be a really small bit of work and might improve our users trouble-shooting experience a lot. Eventually, I'd love to have inheritable error classes but I don't think we need that now.

README.md
lrlna: cool!

lrlna:
> when exceptions have simple names but inherit from classes so you can distinguish what type you have. Not sure we need to implement error inheritance now, but the names look like:

I like simple names! I am down with that!

The question I do want to put out there for both of you is whether or not we want our error codes do be different from the regular javascript errors or something that's similar to what there is in javascript. For example, `TypeError` is what exists in the JS world, and is well understood, so yay! But a nay for that is will the user know right off the bat whether or not the `TypeError` comes from `bson-compilers` or somewhere else in the code? 

(I am kind of torn between one or the other, so definitely something to think about!)

aherlihy:
@lrlna I think it's import to pattern match with real errors, so it's easy for people to understand what's going on. But we should provide our own error class, so that there's no ambiguity that it's invalid input code and not a error being thrown within the compilation code. We could tack on "BSON" to the error name, so "BSONTypeError" instead of "TypeError"?

A separate thought is we could catch _all_ errors within bson-compilers, and then add a "something went wrong, please file a bug report" error as well. Not the same as the generic "RuntimeError" since those still indicate that there was an issue with the input code.

lrlna:
@aherlihy maybe something like `BSONCompilersTypeError` then? still the same `TypeError`, but comes from `BSONCompilers` module? I feel like just `BSON` means there is something wrong with `BSON` not with our compiler.

Yes to `something went wrong`!

aherlihy:
@lrlna I added a commit (b0359953ce06cafadb8a993e5886436f27fe1523) with update to reflect the changes I made in #95

PR #95, opened 2018-06-05
Rename errors and bugfixes

aherlihy:
- Rename errors so that they are simpler and easier to understand - the errors now are named "BsonCompilers" + the error type + "Error". I also updated the messages that are set where we throw the errors, so they are more consistent and more explanatory. We also now have only one type of error, `ArgumentError` for both argument type and count mismatch errors. 

- Rework the error constructor to take in a string since 100% of the time we were calling `new Error({message: "..."})`. The errors take an optional second argument where other attributes can be set like 'line', 'column', 'symbol', in case we want to provide more specific error messages in the future. This is currently just set by ANTLR errors, but that info is not used. Also set the `name` attribute on the errors so that it will be reported as that error in the stack trace, instead of just `Error`.

- Catch all internal errors in index.js and rethrow them as `BsonCompilersInternalError`.

- Update error tests to match the new error changes.

- This PR also includes the fixes I requested in https://github.com/mongodb-js/bson-compilers/pull/85. Those include:
  - Remove the added manual checking for arguments in `Regex`,  `_regex`, `Date`, `LongToString` since it's just repeating code from `checkArguments`.
  - Remove the added try/catch in `Long`, `ObjectId`, `LongToString`, and `Date` (We don't need to catch TypeError and re-throw our own error equivalent in places where we are calling `CheckArguments`, since that is exactly what that method covers.)
  - We don't need both "Reference" and "UndefinedSymbol" errors, they are the same thing.


codegeneration/shell/Visitor.js
```diff
@@ -120,10 +120,10 @@ class Visitor extends JavascriptVisitor {
     const subtype = parseInt(argList.singleExpression()[0].getText(), 10);
     const bindata = args[1];
     if (!(subtype >= 0 && subtype <= 5 || subtype === 128)) {
-      throw new SemanticGenericError({message: 'BinData subtype must be a Number between 0-5 or 128'});
+      throw new BsonCompilersRuntimeError('BinData subtype must be a Number between 0-5 or 128');
```
lrlna: would this be a [BsonCompilersRangeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError) cause it's technically looking for a number in between numbers? or still a Runtime?

index.js
```diff
@@ -60,8 +62,15 @@ const getCompiler = (visitor, generator, symbols) => {
...
-    const tree = loadTree(input);
-    return compiler.start(tree);
+    try {
+      const tree = loadTree(input);
+      return compiler.start(tree);
+    } catch (e) {
+      if (e.code && e.code.includes('BSONCOMPILERS')) {
+        throw e;
+      }
+      throw new BsonCompilersInternalError(e.message, e);
```
lrlna: yesssss!

codegeneration/shell/Visitor.js
aherlihy: Forgot about RangeError! I will add it.

PR #101, opened 2018-06-07
Throw UnimplementedError

aherlihy:
NOTE: needs to be merged after #100.

### DIFF
--- a/codegeneration/javascript/Visitor.js
+++ b/codegeneration/javascript/Visitor.js
@@ -8,7 +8,8 @@ const {
   BsonCompilersRuntimeError,
   BsonCompilersTypeError,
   BsonCompilersReferenceError,
-  BsonCompilersInternalError
+  BsonCompilersInternalError,
+  BsonCompilersUnimplementedError
 } = require('../../helper/error');
 const { singleQuoteStringify } = require('../../helper/format');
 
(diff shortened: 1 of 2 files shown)


codegeneration/javascript/Visitor.js
```diff
@@ -23,6 +24,25 @@ class Visitor extends ECMAScriptVisitor {
     this.new = '';
     this.processInt32 = this.processNumber;
     this.processDouble = this.processNumber;
+
+    // Throw UnimplementedError for nodes with expressions that we don't support
+    this.visitThisExpression =
```
lrlna: are all of these meant to be assigned `this.unimplemented`? if so, i find it a bit unreadable :/

aherlihy: Yeah, all those need to be assigned to the same thing. What do you think is the nicest way to do multiple variable assignment? I thought assigning each on their own line would be too repetitive, but don't mind either way!

lrlna: hmmmm i noticed in #102 there is even more of these, so probably just leaving as you have it is good. I assume at we might have these implemented at some point, so the list will shrink.

aherlihy: Yeah, I thought about having the compilation step make it default to throwing unimplemented errors for _all_ visit methods, and then just override the ones that we don't want to throw in Visitor.js, but then I thought that this list of unimplemented nodes will only get smaller and it'll probably end up being more work in the long run.

PR #105, opened 2018-06-11
Syntax Tests + Fixes

aherlihy:
This PR adds tests for various syntax errors (`test/json/error/javascript/syntax-error.json` for error cases, and `test/json/success/javascript/syntax.json` for valid cases) as well as fixes needed for the tests to pass. The main changes can be summarized with:

1. Add a visit method for equality, so `===` can be transformed into valid target language (`==`). I've also added a new yaml file called `syntax_templates.yaml` where we can specify syntax transformations as templates. (Didn't want to put it in `symbols.yaml` because they should't be searchable in the symbol table, and they're not types so they shouldn't go in `types.yaml`/`basic_types.yaml`). The syntax templates also include options for end-of-statement or end-of-line for future use.
2. Grammar bugfix: previously, the statement `1ObjectId()` would have been valid because it was interpreting `1` as one statement, and `ObjectId()` as another statement. I updated the grammar so that it requires an end-of-line symbol between statements. So now `1;ObjectId()` and `1\\nObjectId()` is valid, but not `1ObjectId`. 
3. Add a few new cases for `UnimplementedError`.
4. Update the tests so that it works for node v9.4. All that really required was removing error tests where we expected invalid `Buffer` or `RegExp` to throw, since they don't throw in 9 but they do in 8.

The last syntax tests-related thing I need to do is add tests for function calls without arguments, so for example `Double`. Technically, that is a valid expression, so I'm going to write some tests to make sure we're treating those types of statements correctly. But I didn't want this PR to get huge so I'm going to do it in the next one.


codegeneration/javascript/Visitor.js
```diff
@@ -68,7 +71,21 @@ class Visitor extends ECMAScriptVisitor {
   }
 
   visitEof() {
-    return '';
+    if (this.Syntax.eof.template) {
+      return this.Syntax.eof.template();
+    }
+    return '\n';
```
lrlna: i am not sure i understand this new line thing going on here? is this for formatting purposes?

aherlihy: If I don't overide the visitEof, then it will generate the string "EOF" at the end of every line, lol. That was a result of moving the starting node from `ExpressionSequence` to `Program`, so that we could handle empty statements.

PR #107, opened 2018-06-14
fix nested documents bugs in c# and python

lrlna:
There was a bug with c# that would add extra `{}` to a larger, more nested query. This PR fixes it, and add a few more tests for nested queries. 

I also noticed that with some queries I would be running, python would add an extra set of `''` on top of `""`. So we would get `'"status"'`, `'"qty"'` where I assume it should be `'status'`, `'qty'`. If my assumption is incorrect, I can undo.


lrlna:
Also! I added three kind of variations I could think of with nested documents, but if y'all feel like we need more, let me know what kinds I should write up!

symbols/csharp/templates.yaml
```diff
@@ -68,25 +68,29 @@ Templates:
...
-            }, '');
+            if (arg.length === 1) {
+              return `new BsonDocument(${doubleStringify(arg[0][0])}, ${arg[0][1]})`
+            }
+
+            const pairs = arg.map((pair) => {
+                return `{ ${doubleStringify(pair[0])}, ${pair[1]} }`;
+            }).join(', ');
+
+            return `new BsonDocument { ${pairs} }`
```
lrlna: there is probably a smarter way of doing this `map` thing. I went with this cause it seemed the easiest / more readable version.

aherlihy: 100% for more readable!

PR #112, opened 2018-06-19
add depth param to help with object/array indents

lrlna:
This PR introduces the concept of `indentDepth` to help us figure out indentation for fomatting purposes. Ideally in the future we'd have this outside of this module, but for now it rests here. 

Indentation is not yet perfect for java arrays. I am not sure how to handle that at the moment, since just separating at comma values yields quite a bizarre result. It's, however, a lot better than what we have currently. 

@aherlihy : would be cool if you take a look at this once you get back from holidays. You might have a better idea how to handle some of the edge cases!


symbols/shell/templates.yaml
```diff
@@ -48,13 +54,20 @@ Templates:
...
+            }
+            return literal;
         }
     ObjectTypeArgsTemplate: &ObjectTypeArgsTemplate !!js/function >
-        (args) => {
-            return args.map((arg) => {
-                return `${arg[0]}: ${arg[1]}`;
+        (args, depth) => {
+            const indent = '\n' + '  '.repeat(depth);
+            const closingIndent = '\n' + '  '.repeat(depth - 1);
```
lrlna: `closingIndent` is to set up an indent for the closing bracket. The closing bracket is giving the same depth as the object it's with, but it needs to have an indent of the previous object. Therefore, I de-indent it by 1. For example:
```js
{
  x: {
    y: 'woot'
  // technically the same depth as y, but should have indentation of x
  }
// technically the same depth as x, but should have indentation of none.
}
```

lrlna:
@aherlihy i am merging this one in, but we can always fix and revert when you're back :sparkles:

codegeneration/javascript/Visitor.js
```diff
@@ -98,9 +98,10 @@ class Visitor extends ECMAScriptVisitor {
    *    separator - a string separator to go between children.
    *    ignore - an array of child indexes to skip.
    *    children - the set of children to visit.
+   * @param {Integer} depth
    * @returns {String}
    */
-  visitChildren(ctx, options) {
+  visitChildren(ctx, options, depth) {
```
aherlihy: I don't think this is needed, since the `getIndentDepth` will skip nodes that don't have depth set.

codegeneration/javascript/Visitor.js
```diff
@@ -121,23 +122,24 @@ class Visitor extends ECMAScriptVisitor {
     return code.trim();
   }
 
-  visitEqualityExpression(ctx) {
+  visitEqualityExpression(ctx, depth) {
```
aherlihy: I also think this doesn't need to be passed down to children.

codegeneration/javascript/Visitor.js
```diff
@@ -121,23 +122,24 @@ class Visitor extends ECMAScriptVisitor {
...
   }
 
   /**
    * Child nodes: literal
    * @param {LiteralExpressionContext} ctx
+   * @param {Integer} depth
    * @return {String}
    */
-  visitLiteralExpression(ctx) {
+  visitLiteralExpression(ctx, depth) {
```
aherlihy: Same comment as the other visit methods.

aherlihy:
I made a new PR #115 with my requested changes, which is just removing the `depth` argument entirely because it's never actually used.

PR #115, opened 2018-06-25
Remove depth argument

aherlihy:
Hi!

The `depth` argument is never used, so I removed it. I thought that I had removed it on my branch before going on vacation, did you re-add it? Or did it not get removed in the first place? Sorry if it didn't get removed, not sure what happened.

(Update: I checked and it looks like I removed `depth` on June 14th, which was the last day I wrote code. The commit that readds it was added on the 15th, so I think it got re-added in a rebasing issue)

Either way, the `getIndentDepth` method handles that, and we can't pass an extra argument to the visit methods without modifying all the generated visit functions, so it was getting lost anyway.

Not sure why the lib files were out of date, I didn't change the templates in the PR but running `npm run compile` on a clean master does make changes to lib.


lrlna:
hi! i pulled the latest changes from that `depth` branch, so i think something happened with your push :O if `getIndnetDepth` handles it, then it's all good! :D

aherlihy:
@lrlna Hi! It looks like the commit that readds it was added on the 15th, when I wasn't working, so I think what happened was I force-pushed on the 14th, but it was still on your local branch so you might have force-pushed on the 15th and re-added it? Def my mistake, I shouldn't force-push to shared branches :)

PR #117, opened 2018-06-27
indent items in csharp arrays

lrlna:
it's not 100% accurate in terms of indents, but makes it a lot more readable.


codegeneration/javascript/Visitor.js
```diff
@@ -207,7 +207,9 @@ class Visitor extends ECMAScriptVisitor {
         return arg !== ',';
       });
       if (ctx.type.argsTemplate) { // NOTE: not currently being used anywhere.
-        args = ctx.type.argsTemplate(visitedElements, ctx.indentDepth);
+        args = visitedElements.map((arg) => {
+          return ctx.type.argsTemplate(arg, ctx.indentDepth);
+        }).join(',');
```
aherlihy: I think we should let the args templates deal with the commas, in case we need to manipulate those as well sometime in the future. Would just change this to `.join()` and then add a comma in the argsTemplate

aherlihy: (I think the ways arrays are structured in C means that we can't assume it's just a comma)

lrlna: that makes sense yea! I do have to do then add a little check on whether it's the last element in the array to pass to the `template` since I don't need a comma there. Let me know if this is okay with you!

lrlna: and [this](https://github.com/mongodb-js/bson-compilers/pull/117/files#diff-496dc8c897880594711d98818f9f01abR211) is what i mean by `last` check.

PR #121, opened 2018-07-27
More granular imports

aherlihy:
This PR changes the API so that instead of exporting a compile function directly, it exports an object with two fields: `compile`, which is the compile function previously being exported directly, and `getImports`, a function that returns a string with the used imports for the code compiled most recently with the `compile` function. `getImports` returns an empty string if nothing has been compiled.

So usage would be:
```
const compiler = require('bson-compilers');
const outputString = compiler.javascript.java.compile("'this is the code to be compiled'")
const exports = compiler.javascript.java.getImports()
```

Changes:
- Add a `code` field to all symbols and types in the symbol table. These codes are unique to the class/type.
- The `npm run compile` command now adds the `symbols/import_templates.yaml` file to the symbol table, so the compiler now has a field `this.Imports` that contains the template (same exact way that `this.Syntax` works)
- Each output language's templates file now has a template for each code, i.e. `0ImportTemplate`. The `ImportTemplate` is the template that combines all the individual import statements.

This will require a major version change since the API change is breaking.


index.js
```diff
@@ -60,66 +60,32 @@ const getCompiler = (visitor, generator, symbols) => {
...
-        idiomatic;
-      return compiler.start(tree);
-    } catch (e) {
-      if (e.code && e.code.includes('BSONCOMPILERS')) {
-        throw e;
+  return {
+    compile: (input, idiomatic) => {
+      try {
+        const tree = loadTree(input);
+        compiler.idiomatic = idiomatic === undefined ?
```
lrlna: i sometimes write this kind of thing as (works with undefined): 
```js
compiler.idiomatic = idiomatic || compiler.idiomatic
```
but i like the more explicit version also o/

(I feel like maybe during a last review this was just a single `|` operator, and I think I said that it wouldn't work, and it still doesn't cause it bit shifts, but a double `||` does the undefined/null check)

but again a more explicit version is also gr88888

PR #126, opened 2018-09-03
Generators never directly access the tree

aherlihy:
This is the last major thing that needed to be done before we can start on adding new visitors/input languages.

This PR adds accessor methods to the visitor so that the generator does not need to call or know anything specific about the tree being traversed. 

For example: previously to get the arguments of a function call, the generator would call `ctx.arguments().argumentsList().singleExpression()`. Now it just calls `this.getArguments(ctx)`, which is defined in the visitor (which returns `ctx.arguments().argumentsList().singleExpression()`).

This is required because when we have multiple trees, the tree access will have different names (or even different shapes) so we want to abstract the data access to the visitor so the generators can be truly input-language agnostic.

As I was going through the code I also found a few places where the lines were very long (esp long ternary statements!) and added newlines between them in the files that I was already working in.

### DIFF
--- a/codegeneration/javascript/Visitor.js
+++ b/codegeneration/javascript/Visitor.js
@@ -951,6 +951,54 @@ class Visitor extends ECMAScriptVisitor {
   processBinary() {
     throw new BsonTranspilersUnimplementedError('Binary type not supported');
   }
+
+  // Getters
+  getExpression(ctx) {
+    return ctx.singleExpression();
+  }
+  getArguments(ctx) {
+    return ctx.arguments().argumentList();
+  }
+  getArgumentAt(ctx, i) {
+    return this.getArguments(ctx).singleExpression()[i];
+  }
+  getList(ctx) {
+    if (!('elementList' in ctx) || !ctx.elementList()) {
+      return [];
+    }
+    return ctx.elementList().singleExpression();
+  }
+  getArray(ctx) {
+    if (!('arrayLiteral' in ctx)) {
+      return false;
+    }
+    return ctx.arrayLiteral();
+  }
+  getObject(ctx) {
+    if (!('objectLiteral' in ctx)) {
+      return false;
+    }
+    return ctx.objectLiteral();
+  }
+  getKeyValueList(ctx) {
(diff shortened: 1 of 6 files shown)


codegeneration/javascript/Visitor.js
```diff
@@ -154,13 +154,17 @@ class Visitor extends ECMAScriptVisitor {
     let code = '';
     for (let i = opts.start; i <= opts.end; i += opts.step) {
       if (opts.ignore.indexOf(i) === -1) {
-        code += this.visit(opts.children[i]) + (i === opts.end ? '' : opts.separator);
+        code += this.visit(
```
alenakhineika: Is the result a string? If yes would be more readable to use a string template:
```
code = `${code} ${this.visit(opts.children[i])} ${(i === opts.end) ? '' : opts.separator}`;
```

codegeneration/javascript/Visitor.js
```diff
@@ -610,10 +624,12 @@ class Visitor extends ECMAScriptVisitor {
       }
       const result = this.castType(expected[i], args[i]);
       if (result === null) {
-        const message = `Argument type mismatch: '${name}' expects types ${expected[i].map((e) => {
-          const id = e && e.id ? e.id : e;
-          return e ? id : '[optional]';
-        })} but got type ${args[i].type.id} for argument at index ${i}`;
+        const message = `Argument type mismatch: '${name}' expects types ${
+          expected[i].map((e) => {
```
alenakhineika: would be nice to move it from string template to separate function and use here just a variable:
```
expected[i].map((e) => {
  const id = e && e.id ? e.id : e;
  return e ? id : '[optional]';
})
```

alenakhineika:
Cool update! Code looks much cleaner! I made just couple of code style comments which are not blockers.

codegeneration/javascript/Visitor.js
aherlihy: fixed!

codegeneration/javascript/Visitor.js
aherlihy: fixed!

PR #130, opened 2019-04-02
Update to bson@4

imlucas:
Noticed we're a few versions back on `bson` while working on agg plugin in storybook and updating to 4 allows for better compatibility in browser.


imlucas:
oof... nope nope nope. didn't see all of the underlying changes to jsbson itself

aherlihy:
I thought we were using whatever the node driver was using. I'm open to updating the version, and the tests to match, but I'm not clear on what the benefits are to updating the bson version are. What would we get out of it?

PR #131, opened 2019-05-05
Update bson module to 4.0.2

durran:
BSON module < 4.0.x is not cloud compatible, and even though the aggregations plugin uses BSON 4.0.2, bson-transpilers was using an older version of the module and needed to be updated.

BSON 4.0.2 has some backwards breaking changes that are not compatible with the current stable Node driver, which is still on BSON 1.1.1 (https://github.com/mongodb-js/mongodb-core/blob/master/package.json#L28). First of all, `Symbol` was renamed to `BSONSymbol`, so which this change will do proper translation with BSON 4.0.2, it will not be current driver compatible Javascript. I think this is ok since no one in JS land should really care about `Symbol` as a BSON type and the type itself has been deprecated for several years. Secondly, `Long` was changed to extend from the long module (https://github.com/dcodeIO/long.js) which now requires that `Long` is always instantiated with `new`.


symbols/java/templates.yaml
```diff
@@ -772,8 +772,14 @@ Templates:
...
         }
-    SymbolSymbolTemplate: &SymbolSymbolTemplate null
-    SymbolSymbolArgsTemplate: &SymbolSymbolArgsTemplate null
+    SymbolSymbolTemplate: &SymbolSymbolTemplate !!js/function >
+        () => {
+            return 'Symbol';
+        }
+    SymbolSymbolArgsTemplate: &SymbolSymbolArgsTemplate !!js/function >
+        (lhs, arg) => {
+            return `(${arg})`;
```
durran: This is the only thing I was suspicious of - thought I wouldn't have needed the params in there.

.npmignore
```diff
@@ -3,3 +3,10 @@
 .travis.yml
 .zuul.yml
 test/
+codegeneration/
```
alenakhineika: Why these repos should be ignored by git?

durran: Not git, npm. We don't need to publish them.

symbols/java/templates.yaml
aherlihy: Do you mean you feel like you should have been able to leave the SymbolSymbolArgsTemplate as null, since it's just the same argument being passed through?

durran: No my assumption was that leaving it as `null` would mean the output would be the same as the input, which needed to change. But I thought I could just do `return arg` and not need the parens, but turned out I did.

.npmignore
alenakhineika: oh i see, sorry for confusing!

symbols/java/templates.yaml
aherlihy: Ah yes, parens are included because sometimes the required translation doesn't take the shape of `x(arg)`, for example for translating the `Double(arg)` into Java, the translation could be `Double.parseDouble("1")`, or `1d`, or `(double) x` depending on the type of the argument.

That being said, since the argument itself doesn't need any translation here, you can actually leave the `SymbolSymbolArgsTemplate` as null because if it's not defined, it will default to `( argument )`. [happens here](https://github.com/mongodb-js/bson-transpilers/blob/master/codegeneration/CodeGenerationVisitor.js#L326-L333).

durran: Ah thanks!

PR #133, opened 2019-09-04
COMPASS-3823: Add git info for license metadata

imlucas:
This PR adds the git URL to package.json so compass can exclude it as a third party dependency when generating licensing information.


imlucas:
(self merging as this is purely a metadata change for our legal department)

PR #140, opened 2021-02-22
docs: move pdf to google docs

mcasimir:


rose-m (approved):
Works, PDF is also publicly visible 👍

PR #141, opened 2021-02-23
feat: compile functions to js

mcasimir:
- Allow functions to be compiled to javascript.
- Throws `Support for exporting functions to languages other than javascript is not yet available.` for other languages.

Other changes:

- Move CI from travis to github workflows

NOTE: Arrow functions are not supported by the grammar. Updating the grammar with a [newer version](https://raw.githubusercontent.com/antlr/grammars-v4/master/javascript/ecmascript/JavaScript/ECMAScript.g4) breaks other features.

### DIFF
--- a/codegeneration/javascript/Visitor.js
+++ b/codegeneration/javascript/Visitor.js
@@ -296,11 +296,6 @@ module.exports = (CodeGenerationVisitor) => class Visitor extends CodeGeneration
    * Process Methods
    *
    */
-  generateFuncDefExpression() {
-    throw new BsonTranspilersUnimplementedError(
-      'Support for exporting functions to languages other than javascript is not yet available.'
-    );
-  }
 
   /* Numerical process methods */
   processNumber(ctx) {
--- a/codegeneration/javascript/Visitor.js
+++ b/codegeneration/javascript/Visitor.js
@@ -28,7 +28,6 @@ module.exports = (CodeGenerationVisitor) => class Visitor extends CodeGeneration
     this.visitTypeofExpression =
     this.visitInExpression =
     this.visitInstanceofExpression =
-    this.visitFuncDefExpression =
     this.visitAssignmentExpression =
     this.visitAssignmentOperatorExpression =
     this.visitMemberIndexExpression =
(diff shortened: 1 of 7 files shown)


codegeneration/javascript/Visitor.js
```diff
@@ -288,11 +287,20 @@ module.exports = (CodeGenerationVisitor) => class Visitor extends CodeGeneration
...
+  visitFuncDefExpression(ctx) {
+    return this.generateFuncDefExpression(ctx);
+  }
+
   /*
    *
    * Process Methods
    *
    */
+  generateFuncDefExpression() {
```
rose-m, line 299: I don't think this should go here. IMHO this should be part of the `CodeGenerationVisitor` that also contains the other `generateXXX` methods. It will then be overridden in the `Generator.js` file as you did.

test/functions.test.js
```diff
@@ -0,0 +1,29 @@
...
+      bsonTranspilers.shell.javascript.compile('function(){}'),
+      'function(){}'
+    );
+  });
+
+  ['object', 'csharp', 'java', 'python'].forEach((language) => {
+    it(`throws an unsupported error compiling functions to ${language}`, () => {
+      assert.throws(
+        () => {
+          bsonTranspilers.shell[language].compile('function(){}');
```
rose-m, line 20: I'm confused this works. As I read it we are transpiling from `shell` to `language` here... So maybe the shell visitor is the JavaScript one and since its a merged inheritance that works. Magic but IMHO an accident 🙈

rose-m (changes requested):
I'd see that default implementation put somewhere else 😅

alenakhineika:
One more comment is that we do not blindly use grammar from `antlr/grammars-v4`, we downloaded one with minimum language support and build on top of it. We can change the grammar itself to add some language features support, or we can use a symbol table for these needs.

rose-m (approved):

alenakhineika (approved):

mcasimir:
>  we do not blindly use grammar from antlr/grammars-v4

Why did we decide to do so? don't we need to maintain all the grammars for all the languages like that?

Anyway, how could we support arrow functions then?

